PR #19, opened 2014-04-10
ENH: allow for custom controller.

cournape:
This simple PR allows for an adapter to take a custom controller.

The rationale is that for legacy reasons, I may have to support etag caching where I ignore the query string (think secured s3 entries). The simplest way seems to have my own controller subclass that passes a different key to the underlying cache


hexagonrecursion:
IMHO merging this was a mistake. While this is a small change it has big implications: it makes the interface of `controller.CacheController` part of the public API and any changes to it are potentially breaking. This use case should have been addressed with something that does not create such a big commitment.

PR #57, opened 2015-01-25
add an extra filecache in setup.py

parkouss:
This allows to use pip with a requirement like 'cachecontrol[filecache]' to automatically install the lockfile dependency. (See https://pip.readthedocs.org/en/1.1/requirements.html#the-requirements-file-format)

Currently I use cache control in one project, and I have something like this in my setup.py:

install_requires = [
    'cachecontrol',
    'lockfile', # this is required for the cache control with file cache use
  ]

This patch will allow me to write instead:
install_requires = [
    'cachecontrol[filecache]',
  ]

What do you think ?


ionrock:
This is great! Thanks for the PR!

PR #58, opened 2015-01-26
Use JSON to serialize the cache instead of pickle

dstufft:
This solves a few issues with the current cache:

* Due to the use of ``pickle.HIGHEST_PROTOCOL`` a cache that is created on Python 3.4 is not able to be loaded on Python 2.7 which can cause things to miss the cache just because of the pickle protocol.
* Pickle itself is an insecure serialization format, by using JSON we'll instead not load files that can execute arbitrary code.
  * In order to support arbitrary binary data this will use base64 encoding to turn them into strings. It has to do some encoding dances to make sure that the ``bytes``/``str `` types are correct. I'm not 100% sure that using ``utf8`` to encode strings that urllib3 has for request headers and the such will work though in all situations (maybe @sigmavirus24 knows?)
* Reduces file size of the cache itself by compressing the cached object.

### DIFF
--- a/cachecontrol/serialize.py
+++ b/cachecontrol/serialize.py
@@ -1,8 +1,27 @@
+import base64
 import io
+import json
+import zlib
 
 from requests.structures import CaseInsensitiveDict
 
 from .compat import HTTPResponse, pickle
+
+
+def _b64_encode_bytes(b):
+    return base64.b64encode(b).decode("ascii")
+
+
+def _b64_encode_str(s):
+    return _b64_encode_bytes(s.encode("utf8"))
+
+
+def _b64_decode_bytes(b):
+    return base64.b64decode(b.encode("ascii"))
+
+
+def _b64_decode_str(s):
+    return _b64_decode_bytes(s).decode("utf8")
 
 
 class Serializer(object):
@@ -16,11 +35,14 @@ class Serializer(object):
 
         data = {
             "response": {
-                "body": body,
-                "headers": response.headers,
+                "body": _b64_encode_bytes(body),
+                "headers": dict(
+                    (_b64_encode_str(k), _b64_encode_str(v))
+                    for k, v in response.headers.items()
+                ),
                 "status": response.status,
                 "version": response.version,
-                "reason": response.reason,
+                "reason": _b64_encode_str(response.reason),
                 "strict": response.strict,
                 "decode_content": response.decode_content,
             },
@@ -34,7 +56,20 @@ class Serializer(object):
                 header = header.strip()
                 data["vary"][header] = request.headers.get(header, None)
 
-        return b"cc=1," + pickle.dumps(data, pickle.HIGHEST_PROTOCOL)
+        # Encode our Vary headers to ensure they can be serialized as JSON
+        data["vary"] = dict(
+            (_b64_encode_str(k), _b64_encode_str(v))
+            for k, v in data["vary"].items()
+        )
+
+        return b",".join([
+            b"cc=2",
+            zlib.compress(
+                json.dumps(
+                    data, separators=(",", ":"), sort_keys=True,
+                ).encode("utf8"),
+            ),
+        ])
 
     def loads(self, request, data):
         # Short circuit if we've been given an empty set of data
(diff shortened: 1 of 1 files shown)


sigmavirus24:
> maybe @sigmavirus24 knows?

What you're doing looks a lot like betamax =P. So maybe I do know.

> In order to support arbitrary binary data this will use base64 encoding to turn them into strings

Yes. Especially with `application/octet-stream` and anything that has a compressed transfer-encoding (e.g., `gzip`'d data), the only way to preserve the response *exactly* when serializing to JSON is to use base-64 encoding. The arbitrary sequences will be corrupted otherwise (when decoding to `utf8` to store as JSON). For headers though, that *should* be unnecessary.

On Py3, `base64` cares about bytes v text. Python 3 also doesn't have symmetric APIs like Python 2 (i.e., Python 3 text types only have encode and bytes types only have decode, where as Py2 each type has both). So for the functions where you call `encode`/`decode` on function input (not on `base64` output), I would just do a `hasattr` check to be 100% safe. (`/me is soooooo paranoid`)

> I'm not 100% sure that using utf8 to encode strings that urllib3 has for request headers and the such will work though in all situations

I haven't had problems with it *yet*. If my memory serves me correctly, all headers are supposed to be latin-1. (I don't think RFCs 7230-7235 change this. I think the hesitancy I have is because the default encoding for response bodies is no longer latin-1. I don't think headers were affected by this. @Lukasa can confirm/deny/slap me upside the head for forgetting.) Transforming a latin-1 string into utf-8 *should* not cause problems. That said, I'm certain there are latin-1 characters that can't properly be represented in utf-8 and that *may* cause corruption. That said, I don't know how plausible it is for a server to send back those bytes. If you want to be super safe, `base64` encoding them should be the best way to preserve the exact bytes.

---

On an unrelated note, I can't wonder if there's a better way for cachecontrol to present the user with serialization options.

What if `cachecontrol.CacheControl` allowed the user to provide a class (or string, or whatever) that specifies the serialization strategy, kind of like hueristics. The default could be "AutoDetect" to preserve backwards compatibility (like now, the strategy seems to be v0, v1, and v2. @dstufft is proposing a v3). Providing better names could help users make informed decisions about what format. And if you make it extensible, users could add other formats (e.g., yaml, etc.) as third party packages.

dstufft:
To be clear, at least in my experiments the places where I'm using ``utf8`` are places where urllib3 was giving us ``str`` not ``bytes`` instances. The utf8 dance there is to turn that ``str`` into a ``bytes`` for base64 (and then back out again). For the request body we have bytes and we just send that directly into base64.

As far as a serialization strategy, I'm not sure that it's generally useful. In what situation can you see the application itself caring about how the serialization itself is handled? Only cache control itself reads or writes these values.

sigmavirus24:
> In what situation can you see the application itself caring about how the serialization itself is handled?

I don't have one in particular, but I wouldn't be terribly surprised if there come along users who care. Besides, it might make the organization of the serialization code a lot nicer (rather than having `_loads_v0`, `_loads_v1`, `_loads_v2`, ..., `_loads_vN` as the library evolves).

ionrock:
@sigmavirus24 I agree that a user should be able to configure a different serialization pattern. The obvious use cases in my mind are:

 1. Something is broken and they need a fix right now! With this PR, which I think is a good idea, having a good workaround for old code would be helpful.
 2. The cache can use details from the responise. Redis, for example, allows setting its own timeout. If the cache store only gets the serialized response, it will have to re-create the response or inspect it to grab the expires header. 

Fortunately, even though it is completely undocumented, we call `self.serializer.dumps` in the controller, so if someone wanted to monkey patch it, they certainly could. Exposing that as an argument in the constructor seems reasonable. 

Also, rather than creating a new API for serializers, it might be better to add a `before_cache_set` that receives the un-serialized response. The return value are the new key, value, *args passed to the set function. That way someone can make a custom serializer (or just reuse the original) while still having access to the args in a way they can create whatever custom cache without having to worry as much about the API breaking over time. 

I'm just thinking aloud, so maybe other have experiences where this hook idea is absolutely horrible :)

dstufft:
I'm unsure of the implications of your comment heh, do you want this PR to switch to something like that, or do you plan to merge this and then do that as an additional enhancement after?

sigmavirus24:
Sorry. I'm +1 on this PR being merged now and architecture decisions being rolled into a separate issue for discussion.

ionrock:
I brought up the use cases to see if anyone thought that it was important to deal with compatibility now, with this patch. If @sigmavirus24 thinks it looks OK in terms of compatibility with requests then I'll plan on merging it. I'll be bumping a minor version number for it to help communicate it could break things.

dstufft:
For what it's worth, this should have no compat problems. It can still load the old pickle data structures and it'll write the new ones. (excluding bugs in my code of course).

ionrock:
@dstufft I've gone ahead an merged this. It would be nice to have a test confirming these cache store upgrades work correctly.

PR #59, opened 2015-01-28
implement heuristic from rfc7234

mblakele:
I was working with a codebase that only sets last-modified and noticed that firefox would cache static resources but cachecontrol wouldn't. Investigating I turned up the last-modified heuristic in rfc7234 which "caches are encouraged to use". I hope it's acceptable and makes sense as default behavior.

This also addresses a problem where stale cache entries lacking an etag would be evicted, because those cached responses wouldn't be available for future requests using If-Modified-Since.

PEP8: ok. There was a preexisting long-line warning on line 257 and I left that alone.

### DIFF
--- a/cachecontrol/controller.py
+++ b/cachecontrol/controller.py
@@ -130,6 +130,17 @@ class CacheController(object):
             if expires is not None:
                 expire_time = calendar.timegm(expires) - date
                 freshness_lifetime = max(0, expire_time)
+        # Fall back on last-modified using heuristic from
+        # http://tools.ietf.org/html/rfc7234#section-4.2.2
+        # Firefox also does something like this per
+        # https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching_FAQ
+        # http://lxr.mozilla.org/mozilla-release/source/netwerk/protocol/http/nsHttpResponseHead.cpp#397
+        # Unlike mozilla we limit this to 24-hr.
+        elif 'last-modified' in headers:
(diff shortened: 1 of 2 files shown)


ionrock:
Thanks for the PR! I have a couple thoughts that I'd like to get your input on.

First off, since this is a heuristic, I'd really like this to be optional. It makes sense for this to be really easy to turn on, as it is obviously a reasonable default if Firefox is doing this. That said, people often have a rough time with caches, especially in a program, when a response is not changing and it should be. Do you see a good way to make this optional in the constructor? For example, the time period chosen when applying the heuristic should probably be configurable. 

With that in mind, it would be great to implement this as an actual [heuristic](http://cachecontrol.readthedocs.org/en/latest/custom_heuristics.html#caching-heuristics). The problem is that I don't know if the current heuristic implementation would support it. The way it works currently is that prior to testing the response to see if it is cacheable, a heuristic can be applied that allows changing the headers accordingly. In this case, it seems that we'd want to set a `Cache-Control` header to `public` and `Expires` to whatever the configured timeout is when we see a `Last-Modified`. Do you think you could convert this into a `WebBrowserHeuristic`?

Let me know your thoughts and if any of this doesn't make sense. The heuristic functionality was added for just this very use case and I am extremely glad you found out what Firefox is doing. My guiding principle here is explicit is better than implicit when it comes to caching.

mblakele:
Closing in favor of #62

PR #61, opened 2015-02-18
Implement heuristic freshness based on Last-Modified

josephw:
Implement the heuristic freshness calculation implied by RFC 7234, and used by many browsers, of ten percent of the resource's age when there is a `Last-Modified` header but no explicit freshness.

### DIFF
--- a/cachecontrol/heuristics.py
+++ b/cachecontrol/heuristics.py
@@ -91,7 +91,7 @@ class LastModifiedHeuristic(BaseHeuristic):
         if 'expires' not in response.headers:
             if 'cache-control' not in response.headers or response.headers['cache-control'] == 'public':
                 # RFC 7231, 6.1
-                if response.status_code in [200, 203, 204, 206, 300, 301, 404, 405, 410, 414, 501]:
+                if response.status in [200, 203, 204, 206, 300, 301, 404, 405, 410, 414, 501]:
                     if 'last-modified' in response.headers:
                         last_modified = parsedate(response.headers['last-modified'])
                         now = datetime.now()
--- a/cachecontrol/heuristics.py
+++ b/cachecontrol/heuristics.py
@@ -78,6 +78,8 @@ class ExpiresAfter(BaseHeuristic):
         tmpl = '110 - Automatically cached for %s. Response might be stale'
         return tmpl % self.delta
 
+ONE_DAY_IN_SECONDS = timedelta(days=1)
+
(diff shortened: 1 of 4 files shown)


cachecontrol/heuristics.py
```diff
@@ -77,3 +78,36 @@ def update_headers(self, response):
...
+    Apply the heuristic suggested by RFC 7234, 4.2.2,
+    using the Last-Modified date, when no explicit
+    freshness is specified.
+    """
+
+    def update_headers(self, response):
+        if 'expires' not in response.headers:
+            if 'cache-control' not in response.headers or response.headers['cache-control'] == 'public':
+                # RFC 7231, 6.1
+                if response.status in [200, 203, 204, 206, 300, 301, 404, 405, 410, 414, 501]:
```
ionrock: Could we move this list of status codes to be an attribute and make it a set? That will make it a tiny bit more performant and then you can give the status codes a better name and potentially remove the comment.

ionrock:
@josephw This is great! Thanks for the PR! I commented on one small bit, but let me know if you can't get to that and I'll merge and take care of it. 

Also, if you can add your name to the CONTRIBUTORS.rst that would be helpful as well as adding a note in the docs/release_notes.rst. I want to be sure you get credit!

Thanks again!

cachecontrol/heuristics.py
```diff
@@ -77,3 +78,36 @@ def update_headers(self, response):
...
+                        return {'expires': datetime_to_header(expires)}
+        return {}
+
+    def warning(self, response):
+        now = datetime.now()
+        date = parsedate(response.headers['date'])
+        current_age = max(timedelta(), now - datetime(*date[:6]))
+
+        if current_age > _ONE_DAY_IN_SECONDS:
+            return '113 - Heuristic Expiration'
```
josephw: There's a bug here. This logic only runs when the resource is fetched, at which point it'll still be heuristically fresh. It needs to run when the resource is fetched from the cache, but the cached headers at that point include the synthetic `Expires`, indistinguishable from a real one.

I can see a few options;

* Unconditionally set the warning.
* Always store the original headers, and re-run `update_headers` when determining freshness
* Store an extra flag with the cached response indicating that a warning should be calculated if necessary.

josephw:
This now has a simpler implementation merged as #62 . I'll open a separate PR for a couple of changes on top of that.

PR #67, opened 2015-03-16
Fix KeyError if cached response has no date header

msabramo:
This fixes errors like the following, which can happen if the cached response has no `Date` header:

    $ pip install pdb
    Collecting pdb
      Using cached pdb-0.1.tar.gz
    Collecting python-gnupg (from pdb)
      Using cached python-gnupg-0.3.7.tar.gz
    Collecting argparse (from pdb)
      Exception:
      Traceback (most recent call last):
        ...
        File "/Users/marca/dev/git-repos/pip/.tox/py27/lib/python2.7/site-packages/pip/download.py", line 859, in _download_http_url
          stream=True,
        File "/Users/marca/dev/git-repos/pip/.tox/py27/lib/python2.7/site-packages/pip/_vendor/requests/sessions.py", line 476, in get
          return self.request('GET', url, **kwargs)
        File "/Users/marca/dev/git-repos/pip/.tox/py27/lib/python2.7/site-packages/pip/download.py", line 367, in request
          return super(PipSession, self).request(method, url, *args, **kwargs)
        File "/Users/marca/dev/git-repos/pip/.tox/py27/lib/python2.7/site-packages/pip/_vendor/requests/sessions.py", line 464, in request
          resp = self.send(prep, **send_kwargs)
        File "/Users/marca/dev/git-repos/pip/.tox/py27/lib/python2.7/site-packages/pip/_vendor/requests/sessions.py", line 576, in send
          r = adapter.send(request, **kwargs)
        File "/Users/marca/dev/git-repos/pip/.tox/py27/lib/python2.7/site-packages/pip/_vendor/cachecontrol/adapter.py", line 36, in send
          cached_response = self.controller.cached_request(request)
        File "/Users/marca/dev/git-repos/pip/.tox/py27/lib/python2.7/site-packages/pip/_vendor/cachecontrol/controller.py", line 110, in cached_request
          parsedate_tz(headers['date'])
        File "/Users/marca/dev/git-repos/pip/.tox/py27/lib/python2.7/site-packages/pip/_vendor/requests/structures.py", line 54, in __getitem__
          return self._store[key.lower()][1]
      KeyError: 'date'

Cc: @alex, @dstufft

### DIFF
--- a/cachecontrol/controller.py
+++ b/cachecontrol/controller.py
@@ -117,6 +117,12 @@ class CacheController(object):
             return resp
 
         headers = CaseInsensitiveDict(resp.headers)
+        if not headers or 'date' not in headers:
+            # With date or etag, the cached response can never be used
+            # and should be deleted.
+            if 'etag' not in headers:
+                self.cache.delete(cache_url)
+            return False
 
         now = time.time()
         date = calendar.timegm(
--- a/tests/test_cache_control.py
+++ b/tests/test_cache_control.py
@@ -128,6 +128,12 @@ class TestCacheControlRequest(object):
     def req(self, headers):
         return self.c.cached_request(Mock(url=self.url, headers=headers))
 
+    def test_cache_request_no_headers(self):
+        cached_resp = Mock(headers={'ETag': 'jfd9094r808', 'Content-Length': 100})
+        self.c.cache = DictCache({self.url: cached_resp})
+        resp = self.req({})
+        assert not resp
+
     def test_cache_request_no_cache(self):
         resp = self.req({'cache-control': 'no-cache'})
         assert not resp


cachecontrol/controller.py
```diff
@@ -117,6 +117,8 @@ def cached_request(self, request):
             return resp
 
         headers = CaseInsensitiveDict(resp.headers)
+        if not headers or 'date' not in headers:
+            return False
```
dstufft: Should this do the same check as down below? Check if there is an ``ETag`` and if there isn't one delete this item from the cache? Without a date or an etag this response will never be used for anything.

dstufft:
This looks good to me, sans the one comment.

cachecontrol/controller.py
msabramo: Yes, good point. Updated.

cachecontrol/controller.py
```diff
@@ -117,6 +117,12 @@ def cached_request(self, request):
             return resp
 
         headers = CaseInsensitiveDict(resp.headers)
+        if not headers or 'date' not in headers:
+            # With date or etag, the cached response can never be used
+            # and should be deleted.
+            if 'etag' not in headers:
```
ionrock: I'm not sure this logic is correct. An Etag should be enough to cache, even if there is not a date because the client can still send a conditional request. I think the test should be `'etag' not in headers and 'date' not in headers`. Does that makes sense? Let me know if I'm misunderstanding.

dstufft: I don't think that's correct.

This method has to return a cached response or ``False``. With the returned value of this, the cachecontrol adapter will either use the returned cached response without hitting the network, or it'll hit the network. Since there's no date header we can never return an object that doesn't have a date header associated with it since we can't determine what the fresh-ness is.

When this returns False the adapter will do some network, if the ETag is fine it'll do the update_cached_response call and load the cached item then outside of this method.

ionrock: Ah ok, I see it now.

PR #68, opened 2015-03-16
Split one line in two; to make pdb easier

msabramo:
With Mock stuff and cached_request on one line, it is difficult to step
into cached_requests without getting stuck in the innards of mock.


ionrock:
Good call. Thanks @msabramo!

PR #69, opened 2015-03-16
Handle TypeError error when reading a Python 2 pickled str in Python 3

msabramo:
e.g.: ` TypeError: 'str' does not support the buffer interface` could happen if cache populated with one version of pickle and read by another version of pickle

For an example, see https://github.com/pypa/pip/issues/2481#issue-59983180

Cc: @alex, @dstufft

### DIFF
--- a/cachecontrol/serialize.py
+++ b/cachecontrol/serialize.py
@@ -126,7 +126,19 @@ class Serializer(object):
             if request.headers.get(header, None) != value:
                 return
 
-        body = io.BytesIO(cached["response"].pop("body"))
+        body_raw = cached["response"].pop("body")
+
+        try:
+            body = io.BytesIO(body_raw)
+        except TypeError:
+            # This can happen if cachecontrol serialized to v1 format (pickle)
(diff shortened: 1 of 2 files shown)


alex:
Personally I'm not a fan of catching such broad exceptions, I'd much rather we handled the bad types in the specific places they occur in the individual deseiralizers.

msabramo:
Yeah, that's a good point. `TypeError` is pretty general and can be raised because of a simple typo.

dstufft:
Agreed with @alex, especially since the deserializers already are handling cases where the data can't be deserialized for whatever reasons.

msabramo:
Updated to handle `TypeError` at a lower level, where it's less likely to mask some other problem.

msabramo:
How does this look now?

msabramo:
Awesome! Do you want to do a release with this and #67? Then pip could vendor in the new release and that would resolve https://github.com/pypa/pip/issues/2481

ionrock:
Done! 0.11.2 is released. Thanks @msabramo!

PR #79, opened 2015-05-20
Make it possible to specify a different type of lock for file cache

dstufft:


ionrock:
@dstufft This has been released with 0.11.5. Thanks!

Issue #81, opened 2015-06-11
Chunked responses are not cached


bdeeney:
Otherwise cacheable chunked responses (indicated by a response header `Transfer-Encoding: chunked`) are not cached by CacheControl when `requests>=2.6.1` is used.

It seems this issue appeared with the inclusion of shazow/urllib3@f21c2a2b73e4256ba2787f8470dbee6872987d2d in the `requests` codebase via kennethreitz/requests@5fcd843eb23e8e84cd5f60c6d372ef4d678f80fe.

Example script demonstrating the issue:

```python
In [1]: import requests
In [2]: from requests.packages import urllib3
In [3]: from cachecontrol import CacheControl
In [4]: urllib3.disable_warnings()  # https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning
In [5]: sess = requests.session()
In [6]: cached_sess = CacheControl(sess)
In [7]: # non-chunked response
In [8]: response = cached_sess.get('http://httpbin.org/cache/60')
In [9]: response.headers.get('transfer-encoding') is None
Out[9]: True
In [10]: response.from_cache
Out[10]: False
In [11]: response = cached_sess.get('http://httpbin.org/cache/60')
In [12]: response.from_cache
Out[12]: True
In [13]: # chunked response
In [14]: response = cached_sess.get('https://tor.eff.org/')
In [15]: response.headers.get('transfer-encoding') == 'chunked'
Out[15]: True
In [16]: response.from_cache
Out[16]: False
In [17]: response = cached_sess.get('https://tor.eff.org/')
In [18]: response.from_cache
Out[18]: False
In [19]:
```

jrydberg:
I'm running into this when trying to cache responses from api.github.com

ionrock:
@jrydberg Can you try @bdeeney's patch to see if that fixes things?

stickystyle:
@ionrock I was experiencing this same issue, using the patch from @bdeeney solved it.

jrydberg:
@ionrock It does.

ionrock:
This should be fixed. Closing this out.

bdeeney:
@ionrock Thanks for identifying a solution for this issue.  Any chance of cutting a pypi release that includes the changes in a10539c?

ionrock:
@bdeeney I'd like to, but I've hoped to get some feedback to ensure it is working as expected. Have you been using the latest ref anywhere?

bdeeney:
[...]

PR #82, opened 2015-06-11
Ensure chunked responses are cacheable

bdeeney:
Proposed patch with covering test case for #81.

Unfortunately, the newly added [urllib3.response.HTTPResponse.read_chunked()](https://github.com/shazow/urllib3/blob/f21c2a2b73e4256ba2787f8470dbee6872987d2d/urllib3/response.py#L419) method exhausts the response stream by calling `readline()` directly on the CallbackFileWrapper's `fp` attribute.  To ensure the callback is processed after the response stream has been exhausted, this patch registers a response hook on the request.  Not a terribly elegant solution, and I worry about the potential for unintended side effects.  I'd welcome any suggestions for modifying/improving this patch.


bdeeney:
I've updated this PR after further testing revealed some issues with the previously submitted patch.  This latest version removes the CallbackFileWrapper class entirely, which I think is necessary in order to support caching of server responses sent with `Transfer-Encoding: chunked`.

tests/test_chunked_response.py
```diff
@@ -0,0 +1,116 @@
...
+    That's your business. But as-is, it works reasonably well for streaming
+    chunked data from a server.
+    """
+    daemon_threads = True
+
+    @property
+    def base_url(self):
+        return 'http://{addr[0]}:{addr[1]}/'.format(addr=self.server_address)
+
+
```
ionrock: Can we move this to the conftest.py and make sure it gets cleaned up when the test suite ends?

ionrock:
@bdeeney This looks reasonable to me. I'm still a little concerned about the hook bit as I've never used them before and am not clear on how reliable that is compared to explicitly wrapping the file handler. I'm also hoping @dstufft or someone else working on pip could give this a try to see if this breaks them.

dstufft:
I think this is going to end up with the problem that the original callback file wrapper attempted to solve-- namely that we wanted to delay actually caching the response until the calling application itself read all of the data. In pip this is important in order to make the progress reporting work. I do not know for a fact that this would break that though.

ionrock:
@bdeeney So, I've gotten around to testing this out and @dstufft is correct that it breaks the use case where the response body is cached even when the response hasn't been used. I'm working on @rmcgibbo's patch #106 instead.

Issue #88, opened 2015-07-29
CacheControl fails on invalid (non-integer) max-age header


hakanw:
CacheControl totally fails on this URL:

'http://www.meristation.com/v3/podcasts_rss.php'

Due to the max-age header being '600s' which is obviously wrong, but I think CacheControl should catch the ValueError (trying to cast the string to an int) and handle it gracefully.

Exception:

``` python
ValueError
invalid literal for int() with base 10: '600s'
Location: cachecontrol/controller.py, cache_response:241
```

sigmavirus24:
I agree. That said, CacheControl *probably* should raise an error of its own to let you know that there were cache headers present, but they were invalid. You might otherwise expect that this was cached if you inspected the response and missed that header being malformed.

hakanw:
Maybe a warning or log an error? But to die completely I think is over the top. I use cachecontrol as a wrapper around all my HTTP handling, so I expect it to be solid and withstand all the weird stuff on the web. ;)

sigmavirus24:
Yeah, I guess its unreasonable for anyone talking to the internet to handle any failures at all

hakanw:
Haha, I meant the opposite of course. But yes, handling the internet is a pretty big task.

ionrock:
@hakanw Thanks for catching this one! I'll take a look.

woodruffw:
Long fixed, I believe.

PR #89, opened 2015-07-31
Detect if headers are bytes/unicode when encoding

sigmavirus24:
When base 64 encoding headers, we should be able to smartly detect if
they're already bytes or a `text_type` in `_b64_encode`. This will avoid the
`UnicodeDecodeErrors` in #88 but this doesn't handle the problem where
parts of the Cache-Control information include unicode.

Normalization of those values is left to a different change.

### DIFF
--- a/cachecontrol/compat.py
+++ b/cachecontrol/compat.py
@@ -21,3 +21,9 @@ try:
     from requests.packages.urllib3.util import is_fp_closed
 except ImportError:
     from urllib3.util import is_fp_closed
+
+# Replicate some six behaviour
+try:
+    text_type = (unicode,)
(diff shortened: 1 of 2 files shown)


sigmavirus24:
@ionrock I wasn't sure what the best way was to write a test for this bug. Advice is very welcome.

ionrock:
@sigmavirus24 I added some tests. Thanks so much for looking into this!

sigmavirus24:
Woo my first contribution to cachecontrol! That said, I don't think this quite fixes #88. You still have unicode open/close quotes in the header value and that will probably cause other problems, no?

Issue #92, opened 2015-08-05
CacheControl should log its decisionmaking


toolforger:
All the decisions that lead to CacheControl entering something into the cache or not, retrieving something from the cache or not, should be logged at DEBUG level.
Reason: HTTP caching is far more complex and influenced by far more factors than most developers that will use CacheControl are aware. If developers are searching for the cause of unexpected behaviour (a.k.a. a bug), knowledge of the decisionmaking will help them determine whether they should look in their own logic, the server's logic, or CacheControl's logic.

ionrock:
@toolforger I believe I merged your logging additions. Can we close this one?

toolforger:
Do you want to guarantee to cachecontrol users that all decision points are logged?
It would be very valuable to them. In that case, you'd want to (a) check that I didn't overlook any decision points (I tried to be thorough but can't be sure), and (b) document that guarantee in the docs so users will actually start to rely on it.

Whether you accept that or not, I'm fine with this one being closed, my work is done unless you have questions :-)

PR #93, opened 2015-08-05
Adding logging (issue #92)

toolforger:
First commit is a toe in the water: cache bypass decisions are logged.

PR created so we can discuss whether that's the right direction to take.
I have no idea about testing this. I'm sceptical about testing logging, because it tends to be both volatile and inconsequential so the gains of testing are marginal, but I'm happy to follow whatever the policy is.

### DIFF
--- a/cachecontrol/controller.py
+++ b/cachecontrol/controller.py
@@ -107,7 +107,7 @@ class CacheController(object):
             return False
 
         # Check whether it can be deserialized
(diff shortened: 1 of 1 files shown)


toolforger:
Right, that `global` statement will go. (Not an error actually, but it's gratuitious.)

ionrock:
@toolforger Thanks for the PR. I merged this in the branch for #100. It should be merged into master when the `doesitcache` tool is finished.

PR #98, opened 2015-08-26
Check for valid max-age in responses

josephw:
Motivated by seeing:

    Cache-Control: max-age=900; public

(with a semicolon, rather than a comma) in the wild. It's invalid, but shouldn't cause an error. Right now, this fails with an exception. Fix it, and add a corresponding test when retrieving from the cache, to confirm that that case already checks.

### DIFF
--- a/cachecontrol/controller.py
+++ b/cachecontrol/controller.py
@@ -280,7 +280,7 @@ class CacheController(object):
         elif 'date' in response_headers:
             # cache when there is a max-age > 0
             if cc and cc.get('max-age'):
-                if int(cc['max-age']) > 0:
+                if cc['max-age'].isdigit() and int(cc['max-age']) > 0:
                     logger.debug('Caching b/c date exists and max-age > 0')
                     self.cache.set(
                         cache_url,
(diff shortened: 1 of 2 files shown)


ionrock:
@josephw Sorry for the slow response on this. Would you mind rebasing? 

Also, as this seems like a reasonable thing to work around, that you leave a comment in the tests that this should potentially be fixed? If you have an example of it in the wild, that would be helpful too. 

Thanks and apologies again for taking a while to review this!

josephw:
> Would you mind rebasing?

Sure; rebased, and comments added to the tests. (Do you prefer resolving conflicts in PRs through tebases, or merging `master` into topic branches?)

> If you have an example of it in the wild, that would be helpful too.

True story: for the case where I ran into this, I opened this PR and fired off an email to them at the same time. They got their server fixed first, so I don't have a current instance on the web. I don't think it's a common mistake at all, but it seems like a general improvement in robustness.

ionrock:
@josephw I'm a fan of rebasing, but however you choose to update the PR is up to you. I'm far from a git guru!

Issue #95, opened 2015-08-06
Not getting the caching I expected


toolforger:
Here's my CCSSE:

```
import github3
import cachecontrol
g = github3.GitHub()
cachecontrol.CacheControl(g._session)
print(g.rate_limit()['resources']['core']) # initial rate limit
print(g.rate_limit()['resources']['core']) # rate_limit - should not count
repository = g.repository('sigmavirus24', 'github3.py')
print(g.rate_limit()['resources']['core']) # get repo data - should count
repository = g.repository('sigmavirus24', 'github3.py')
print(g.rate_limit()['resources']['core']) # get repo again - should be served from cache
```

In the output, I'm seeing that the rate limit is being ticked off for each `g.repository` call.

```
{u'reset': 1438863563, u'limit': 60, u'remaining': 53}
{u'reset': 1438863563, u'limit': 60, u'remaining': 53}
{u'reset': 1438863563, u'limit': 60, u'remaining': 52}
{u'reset': 1438863563, u'limit': 60, u'remaining': 51}
```

With logging cranked up to the max and tons of logging added (see #93 for the exact code, 50a76aa to be specific), I'm seeing the long trace below.
Conspicuously, there's no `Updating cache with response from "http://..."`, which is what I added at https://github.com/toolforger/cachecontrol/blob/50a76aa0f022a47d34c65c13c4c813ecb1f2c086/cachecontrol/controller.py#L228, so I guess  indicating  `cachecontrol.controller.CacheController.cache_response` is never called.
Since that call is stashed away in a `functools.partial`, I have no idea where and when that call should have happened, so I have come to a dead end.

```
INFO:github3:Building a url from ('https://api.github.com', 'rate_limit')
INFO:github3:Missed the cache building the url
DEBUG:github3:GET https://api.github.com/rate_limit with {}
DEBUG:cachecontrol.controller:Looking up "https://api.github.com/rate_limit" in the cache
DEBUG:cachecontrol.controller:No cache entry available
INFO:requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): api.github.com
DEBUG:requests.packages.urllib3.connectionpool:"GET /rate_limit HTTP/1.1" 200 None
INFO:github3:Attempting to get JSON information from a Response with status code 200 expecting 200
INFO:github3:JSON was returned
INFO:github3:Building a url from ('https://api.github.com', 'rate_limit')
DEBUG:github3:GET https://api.github.com/rate_limit with {}
[...]

PR #100, opened 2015-09-03
Add a `doesitcache` command line tool

ionrock:
The `doesitcache` tool allows you check a URL to see if it should get
cached or not. Logging is printed to help determine what might be
stopping a response from being cached.

The hope is to help better support things like issue #95.


sigmavirus24:
I would make the cmd module private to prevent people from importing and using it in code, e.g., `cachecontrol._cmd`

rmcgibbo:
Just my 2cents, but it might be preferable to put this in `__main__` to enable `python -m cachecontrol` instead of an extra command line tool.

PR #104, opened 2015-11-23
Update custom_heuristics.rst

rmcgibbo:
By trial and error, it seems that you need to return the headers to be updated.


ionrock:
@rmcgibbo You're correct! Thanks for fixing this in the docs!

rmcgibbo:
np.

PR #106, opened 2015-11-24
Implementing caching of chunked/streamed responses

### LINKED ISSUE #105: Chunked responses are not cached
When the underlying `urllib3.HTTPResponse` is chunked, I don't get any caching.

I believe the issue is coming from the fact that `requests` gets the data out using `urllib3.HTTPResponse.stream()`, which, if the response is chunked, calls `urllib3.HTTPResponse.read_chunked`. This avoids directly hitting the `read()` method on the file object (it calls `_safe_read` instead) that cachecontrol overrides in the `CallbackFileWrapper`.

I'm using python 35 and requests 2.8.1.

rmcgibbo:
Resolves #105, #81. This takes a different approach from #82 in that it doesn't get rid of `CallbackFileWrapper`, so there is still the same delay actually caching the response until the calling application itself reads the data.

On the other hand, this requires dynamically monkeypatching urllib3, so ¯\_(ツ)_/¯.

### DIFF
--- a/cachecontrol/serialize.py
+++ b/cachecontrol/serialize.py
@@ -134,6 +134,12 @@ class Serializer(object):
 
         body_raw = cached["response"].pop("body")
 
+        headers = CaseInsensitiveDict(data=cached['response']['headers'])
+        if headers.get('transfer-encoding', '') == 'chunked':
+            headers.pop('transfer-encoding')
+
+        cached['response']['headers'] = headers
+
         try:
             body = io.BytesIO(body_raw)
         except TypeError:
--- a/tests/test_stream.py
+++ b/tests/test_stream.py
@@ -1,0 +1,23 @@
+"""
+Test for supporting streamed responses (Transfer-Encoding: chunked)
+"""
+import requests
+
+from cachecontrol import CacheControl
+
+
+class TestStream(object):
+
+    def setup(self):
+        self.sess = CacheControl(requests.Session())
+
+    def test_stream_is_cached(self, url):
+        resp_1 = self.sess.get(url + 'stream')
+        content_1 = resp_1.content
+
+        resp_2 = self.sess.get(url + 'stream')
+        content_2  = resp_1.content
+
+        assert not resp_1.from_cache
+        assert resp_2.from_cache
+        assert content_1 == content_2
--- a/tests/test_stream.py
+++ b/tests/test_stream.py
@@ -7,16 +7,22 @@ from cachecontrol import CacheControl
 
 
 class TestStream(object):
-    def setup(self):
-        self.sess = CacheControl(requests.Session())
-
     def test_stream_is_cached(self, url):
-        resp_1 = self.sess.get(url + 'stream')
+        sess = CacheControl(requests.Session())
+
+        resp_1 = sess.get(url + 'stream')
         content_1 = resp_1.content
 
-        resp_2 = self.sess.get(url + 'stream')
+        resp_2 = sess.get(url + 'stream')
         content_2 = resp_1.content
 
         assert not resp_1.from_cache
         assert resp_2.from_cache
         assert content_1 == content_2
+
+    def test_stream_is_not_cached_when_content_is_not_read(self, url):
+        sess = CacheControl(requests.Session())
+        sess.get(url + 'stream', stream=True)
+        resp = sess.get(url + 'stream', stream=True)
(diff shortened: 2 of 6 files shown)


rmcgibbo:
Fixes https://github.com/ionrock/cachecontrol/issues/95#issuecomment-138002461

tests/test_stream.py
```diff
@@ -0,0 +1,22 @@
...
+class TestStream(object):
+    def setup(self):
+        self.sess = CacheControl(requests.Session())
+
+    def test_stream_is_cached(self, url):
+        resp_1 = self.sess.get(url + 'stream')
+        content_1 = resp_1.content
+
+        resp_2 = self.sess.get(url + 'stream')
+        content_2  = resp_1.content
```
ionrock: There are two spaces before the `=` here.

ionrock:
@rmcgibbo In trying this out, I added a test:

    def test_stream_is_not_cached_when_content_is_not_read(self, url):
        self.sess.get(url + 'stream')
        resp = self.sess.get(url + 'stream')
        assert not resp.from_cache

This fails, when (I think) it should pass. I'm thinking a chunked response is essentially the same as if you did something like `sess.get(url, stream=True)`, which means you need to consume the file handle in order to cache it. 

Let me know your thoughts and if you have ideas for fixing it. At the moment, I think it would be most helpful if urllib3 provided a "consumed" method on the HTTPResponse that helped solve this type of problem where CacheControl has to guess that the file handle has been read.

rmcgibbo:
Is that a fresh session or in the TestStream?

ionrock:
@rmcgibbo That is the same session in the TestStream.

rmcgibbo:
Then it depends on the order in which the two tests get executed, no? If your test is executed afterwards, then should be cached.

rmcgibbo:
Or, maybe not. I just added your test in a way that I thought should pass, but no. I'll take a look.

rmcgibbo:
> Let me know your thoughts and if you have ideas for fixing it. At the moment, I think it would be most helpful if urllib3 provided a "consumed" method on the HTTPResponse that helped solve this type of problem where CacheControl has to guess that the file handle has been read.

I'm taking a look now. A consumed method would definitely be cleaner.

rmcgibbo:
@ionrock: this test is failing, even without anything to do with streamed or chunked responses:

```
class TestNonchunked(object):
    def test_not_cached_when_content_is_not_read(self, url):
        sess = CacheControl(requests.Session())
        sess.get(url)
        resp = sess.get(url)

        assert not resp.from_cache
```

https://travis-ci.org/ionrock/cachecontrol/jobs/93289665#L153-L154

scollinson:
Any movement on this PR? Would be nice to use this lib for services with chunked responses.

rmcgibbo:
@scollinson I don't really remember. I haven't looked at this in a long time. But I think this PR is good to go. IIRC, the apparent bug that @ionrock noted above is not a bug in this PR, but actually a bug somewhere else that exists even in the current version of cachecontrol without this PR (see the travis output in https://github.com/ionrock/cachecontrol/pull/107).

cachecontrol/serialize.py
```diff
@@ -134,6 +134,12 @@ def prepare_response(self, request, cached):
 
         body_raw = cached["response"].pop("body")
 
+        headers = CaseInsensitiveDict(data=cached['response']['headers'])
+        if headers.get('transfer-encoding', '') == 'chunked':
+            headers.pop('transfer-encoding')
+
+        cached['response']['headers'] = headers
+
```
ionrock: Is there a good reason for this? It seems like it doesn't hurt to know the original response was chunked.

rmcgibbo: I don't really remember this PR. I think perhaps when the initial request is transferred with chunks, one of the first steps is to strip off the chunk markers, so that the data that actually enters the cache lacks the chunk markers. when it's read back out of the cache then, it doesn't look like a chunked request.

ionrock: Ah ok, that is what I suspected. Thanks!

I've finally gotten around to getting this merged with some changes. Thanks for your patience!

ionrock:
Made some changes and merged this locally. Unfortunately, it seems I blew it and it didn't pick up this PR as merged :(

Closing it now.

PR #107, opened 2015-11-26
Add test_request_not_cached_when_content_is_not_read

rmcgibbo:
cc https://github.com/ionrock/cachecontrol/pull/106#issuecomment-159775663


rmcgibbo:
I believe the root cause of this is [here in requests](https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L607-L608).

ionrock:
This has been added.

PR #112, opened 2016-01-19
Don't cache incomplete downloads

### LINKED ISSUE #83: Detect Failed Downloads
Currently (I think) the caching code assumes that if we call ``read()`` on the file and it's closed that we've successfully downloaded the file. This caused issues for someone in pip (see https://bitbucket.org/zzzeek/sqlalchemy/issue/3447/unable-to-install-sqlalchemy). I'm not sure what the best way of handling this would be.

dstufft:
Fixes #83


ionrock:
Thanks @dstufft!

PR #113, opened 2016-01-19
Add a hook to determine if a response should be cached

dstufft:
This is primarily useful to add extra logic to _prevent_ the caching of a response, but the way it is implemented it can also be used to add extra logic to _force_ the caching of a response without adjusting the response itself.

In addition this also adjusts the default heuristic a bit so that it only applies the warning header if it's actually going to modify the headers, and not unconditionally. This makes it possible to do a simple thing like:

``` python
import random

from cachecontrol.heuristic import BaseHeuristic

class RandomlyCache(BaseHeuristic):
    def should_cache(self, request, response, body):
        return random.choice([True, False, None])
```

Without having it adjust the content of the request or response at all.

### DIFF
--- a/cachecontrol/adapter.py
+++ b/cachecontrol/adapter.py
@@ -25,6 +25,7 @@ class CacheControlAdapter(HTTPAdapter):
             self.cache,
             cache_etags=cache_etags,
             serializer=serializer,
+            heuristic=self.heuristic,
         )
 
     def send(self, request, **kw):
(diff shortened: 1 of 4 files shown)


ionrock:
@dstufft So, in thinking about things a bit, I think that for the time being, it is probably better to avoid hidden state by adding the `should_cache` method to the heuristic. While I do think the heuristics are somewhat low level, the nice thing about them is that they do require you to update the response and work within the confines of normal caching. 

You mentioned in IRC that you can use the heuristic as-is, so I'd love to merge the warning fix in https://github.com/dstufft/cachecontrol/commit/47ad491de4a239575f9f336fe8ca65b42d50a675 but hold off on the `should_cache` method.

PR #114, opened 2016-01-19
[WIP/PoC] Attempt to use the fasteners library

dstufft:


cachecontrol/caches/file_cache.py
```diff
@@ -79,13 +100,21 @@ def _fn(self, name):
         parts = list(hashed[:5]) + [hashed]
         return os.path.join(self.directory, *parts)
 
+    def _lockname(self, name):
+        if isinstance(self.lock_class, PYLockFileBase):
+            return self.name
```
dsully: This should be 'name', not 'self.name'

cachecontrol/caches/file_cache.py
```diff
@@ -79,13 +100,21 @@ def _fn(self, name):
...
+        else:
+            return self.name + ".lock"
```
dsully: Same here.

cachecontrol/caches/file_cache.py
```diff
@@ -79,13 +100,21 @@ def _fn(self, name):
...
+
     def get(self, key):
         name = self._fn(key)
-        if not os.path.exists(name):
-            return None
 
-        with open(name, 'rb') as fh:
-            return fh.read()
+        with self.lock_class(self._lockname(name)):
+            if not os.path.exists(name):
```
dsully: Should this be refactored as a try/except block, instead of testing for path existence?

kloczek:
To be able test that PR it would be necessary to modify module metadata to statr require on install time `fasteners` instead `lockfile`.
Is it possible to add that part ot the PR as well?

Issue #91, opened 2015-08-05
UnicodeDecodeError on some utf8 content in headers in cachecontrol


hakanw:
Hi again, I found another similar case like #84 but not exactly the same place.

URL that is failing: http://wizard2.sbs.co.kr/w3/podcast/V0000372136.xml

Specifically, the decoding chokes on the unicode chars in this header:

`... 'p3p': "CP='\xb0\xa3\xb7\xab\xb9\xe6\xc4\xa7\xb1\xe2\xc8\xa3'" ...`

Stacktrace:

``` python
  File "/home/ubuntu/.virtualenvs/webserver/local/lib/python2.7/site-packages/requests/sessions.py", line 476, in get
    return self.request('GET', url, **kwargs)
  File "/home/ubuntu/.virtualenvs/webserver/local/lib/python2.7/site-packages/opbeat/instrumentation/packages/base.py", line 63, in __call__
    args, kwargs)
  File "/home/ubuntu/.virtualenvs/webserver/local/lib/python2.7/site-packages/opbeat/instrumentation/packages/base.py", line 214, in call_if_samp
ling
    return wrapped(*args, **kwargs)
  File "/home/ubuntu/.virtualenvs/webserver/local/lib/python2.7/site-packages/requests/sessions.py", line 464, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/ubuntu/.virtualenvs/webserver/local/lib/python2.7/site-packages/requests/sessions.py", line 576, in send
    r = adapter.send(request, **kwargs)
  File "/home/ubuntu/.virtualenvs/webserver/src/cachecontrol-master/cachecontrol/adapter.py", line 36, in send
    cached_response = self.controller.cached_request(request)
  File "/home/ubuntu/.virtualenvs/webserver/src/cachecontrol-master/cachecontrol/controller.py", line 102, in cached_request
    resp = self.serializer.loads(request, self.cache.get(cache_url))
  File "/home/ubuntu/.virtualenvs/webserver/src/cachecontrol-master/cachecontrol/serialize.py", line 114, in loads
    return getattr(self, "_loads_v{0}".format(ver))(request, data)
  File "/home/ubuntu/.virtualenvs/webserver/src/cachecontrol-master/cachecontrol/serialize.py", line 180, in _loads_v2
    for k, v in cached["response"]["headers"].items()
[...]

Issue #84, opened 2015-06-29
UnicodeDecodeError raised on some cache max-age headers


hakanw:
I'm encountering this exception when fetching some URLs:

```
UnicodeDecodeError
 'ascii' codec can't decode byte 0xe2 in position 0: ordinal not in range(128)
```

The function that is raising is this:

```python
def _b64_encode_str(s):
    return _b64_encode_bytes(s.encode("utf8"))
```

And some example data that some HTTP servers seem to be sending is for example: `'\u201cmax-age=31536000\u2033'`

It would be great if cache-control could handle these edge cases without dying.

Thanks for a great package!

sigmavirus24:
The code you provided does not exist in cachecontrol.

Could you provide the full header?

sigmavirus24:
Also, what version of Python are you using?

ionrock:
@sigmavirus24 I think @hakanw is looking at https://github.com/ionrock/cachecontrol/blob/master/cachecontrol/serialize.py#L15

@hakanw Is that a public URL you are accessing? It is odd to me that a server is returning headers with opening and closing quotes rather than double quotes.

ionrock:
@hakanw Thanks for creating the issue btw!

hakanw:
I'm running python 2.7.6 and cachecontrol 0.11.5.

Yes, exactly it's that code that is failing!

Yeah I know it's odd but you know, it's the web :) Lots of strange HTTP servers and clients are out there...   I'm writing a spider so I'm running into a lot of these issues.

sigmavirus24:
> @sigmavirus24 I think @hakanw is looking at https://github.com/ionrock/cachecontrol/blob/master/cachecontrol/serialize.py#L15

Ah, GitHub's search failed at finding that.

hakanw:
This is still happening a bunch. Maybe there's some specific server that emits this particular kind of cache header. I just wished cache control didn't totally fail with an exception.

Screenshot from the stacktrace and my logging software:

<img width="636" alt="skarmavbild 2015-07-29 kl 15 29 36" src="https://cloud.githubusercontent.com/assets/100824/8958711/aa518402-3606-11e5-9bd3-0cd6fffc5235.png">

sigmavirus24:
So on Python 2.7.9, if I do

```py
>>> uniheader = u'\u201cmax-age=31536000\u2033'
>>> strheader =  '\u201cmax-age=31536000\u2033'
[...]

PR #115, opened 2016-02-27
Use msgpack for cache serialization

StephanErb:
This is a pull request meant to improve the efficiency of wheel download caching in pip (https://github.com/pypa/pip/issues/3515).

Msgpack is fast, supports all major Python versions, and does not add overhead for the serialization of large binary values (as commonly handled by pip).

**Benchmark results**: 

```
# Before
$ python ./examples/benchmark.py
Total time for 1000 requests: 0:00:00.670020
```

```
# After
$ python ./examples/benchmark.py
Total time for 1000 requests: 0:00:00.574051
```

### DIFF
--- a/cachecontrol/serialize.py
+++ b/cachecontrol/serialize.py
@@ -3,23 +3,10 @@ import io
 import json
 import zlib
 
+import msgpack
 from requests.structures import CaseInsensitiveDict
 
-from .compat import HTTPResponse, pickle, text_type
-
-
-def _b64_encode_bytes(b):
-    return base64.b64encode(b).decode("ascii")
-
-
-def _b64_encode_str(s):
-    return _b64_encode_bytes(s.encode("utf8"))
-
-
-def _b64_encode(s):
-    if isinstance(s, text_type):
-        return _b64_encode_str(s)
-    return _b64_encode_bytes(s)
+from .compat import HTTPResponse, pickle
 
 
 def _b64_decode_bytes(b):
@@ -52,14 +39,11 @@ class Serializer(object):
 
         data = {
             "response": {
-                "body": _b64_encode_bytes(body),
-                "headers": dict(
-                    (_b64_encode(k), _b64_encode(v))
-                    for k, v in response.headers.items()
-                ),
+                "body": body,
+                "headers": dict(response.headers),
                 "status": response.status,
                 "version": response.version,
-                "reason": _b64_encode_str(response.reason),
+                "reason": response.reason,
                 "strict": response.strict,
                 "decode_content": response.decode_content,
             },
@@ -73,20 +57,7 @@ class Serializer(object):
                 header = header.strip()
                 data["vary"][header] = request.headers.get(header, None)
 
-        # Encode our Vary headers to ensure they can be serialized as JSON
-        data["vary"] = dict(
-            (_b64_encode(k), _b64_encode(v) if v is not None else v)
-            for k, v in data["vary"].items()
-        )
-
-        return b",".join([
-            b"cc=2",
-            zlib.compress(
-                json.dumps(
-                    data, separators=(",", ":"), sort_keys=True,
-                ).encode("utf8"),
-            ),
-        ])
+        return b",".join([b"cc=3", msgpack.dumps(data, use_bin_type=True)])
 
     def loads(self, request, data):
         # Short circuit if we've been given an empty set of data
@@ -192,5 +163,13 @@ class Serializer(object):
             (_b64_decode_str(k), _b64_decode_str(v) if v is not None else v)
             for k, v in cached["vary"].items()
         )
+
+        return self.prepare_response(request, cached)
+
+    def _loads_v3(self, request, data):
+        try:
+            cached = msgpack.loads(data, encoding='utf-8')
+        except ValueError:
+            return
 
         return self.prepare_response(request, cached)
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -1,11 +1,10 @@
+import msgpack
 import requests
 
-from mock import Mock, patch
+from mock import Mock
 
 from cachecontrol.compat import pickle
 from cachecontrol.serialize import Serializer
-from cachecontrol.serialize import _b64_encode
-from cachecontrol.serialize import _b64_decode_str
 
 
 class TestSerializer(object):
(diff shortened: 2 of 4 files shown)


StephanErb:
The code-quality check seems to have failed due to a internal server error at their end:


> Check number 12 has failed. This is almost always due to a bug in Landscape - we have been notified. Sorry about that.
> Failure reason: Unknown
> We don't know exactly why this error happened yet

StephanErb:
Anything missing here or things you would want me to address?

/cc @dstufft

ionrock:
@StephanErb sorry the slow response. My only concern is adding the dependency. Right now, CacheControl only requires requests and this adds another dependency that is arguably doesn't add a huge advantage in the general case. 

What do you think about making it optional if msgpack is installed, falling back to the default encoding scheme otherwise? I realize this could get weird if you had more than one process using with msgpack installed and another without, so please take this as a desire to discuss the issue and not a demand! 

Let me know what you think!

dstufft:
It does have *some* advantages, even in the general case:

* The code to serialize is simpler since msgpack has a binary type and JSON does not. This would probably fix #91 and would have prevented #84.
* It's faster that JSON in both small and large messages, making it a speed boost in both scenarios.
* It's smaller, making it so that the cache will consume less space on disk (or in memory, or wherever it is stored).

Of course, it comes with the downside of adding another mandatory dependency (in this case, a dependency that does have an optional C extension for CPython speedups, but it fails gracefully if that fails to compile for any reason).

All in all, my personal opinion is that I think msgpack is great, and I think it makes sense to do this, however if you would prefer not to force it I would suggest that it's likely to be a better idea to have some sort of flag to force this on/off rather than auto detecting (though I'm not sure how to handle that exactly, the versioned cache serialization format doesn't really handle downgrades or "features", just upgrades).

StephanErb:
When preparing the pull request, I also pondered whether to keep the non-msgpack fallback or not.

I concluded that the primary concern of  cachecontrol is caching of HTTP responses. Having a dependency on a first-class serialization library therefore did not seem unreasonable.

StephanErb:
Have you put any more thought into this? If deemed necessary, I'd be happy to update the pull request with a fallback mode to the old serialization format.

ionrock:
@StephanErb Actually yes, I think it would be best to avoid the extra dependency simply b/c people do have hangups about that sort of thing. I'm honestly not one of them, but the fact that msgpack can include C deps makes it more confusing and cumbersome for folks. I can imagine new users seeing a message it can't compile something in some output would feel they are doing something wrong. Even though it doesn't make a difference, it is one of those things that makes it harder getting started. I've always been a fan of CherryPy, partially for this reason. 

With that in mind, as I'm sure this is helpful for pip, if you and @dstufft wanted to pull the file cache bits out into its own library, that doesn't bother me. I've always avoided CacheControl including a ton of storage backends b/c they are difficult to maintain. The file cache was the one exception, but if it seems like it needs to be pulled out into its own library in order to ensure yours truly is not a bottleneck, I'm not completely opposed to the idea. 

Lastly, if it is a total pain in the neck to have a fallback path, I'm curious how the benchmark looks with msgpack using C vs. the pure Python msgpack. If the pure Python version is slower, I'm thinking it would be best to use an explicit method of including msgpack:

    # this completely made up code
    import cachecontrol.config
    cachecontrol.config.use_msgpack = True

That way we help move folks towards the best solution depending on their requirements (ie no C runtime to build in deployment). 

Thanks for your patience and I hope I'm not putting any undue burden on you. Part of me just wants to merge it, but, at the same time, after dealing with deployment issues as of late, adding an extra dependency that does have the potential of compiling C does make me hesitant.

dstufft:
I don't have a problem using another library that extends cachecontrol, but this isn't something that the storage backend controls, this is the serializer which currently is storage agnostic. The serializer appears to be pluggable (can be passed into the controller as an argument). Assuming that is part of the API and isn't just something for testing I wouldn't be opposed to depending on some other library that implements a msgpack serializer for CacheControl (or just doing it in pip in that case I guess).

I do want to point out that all modern versions of pip will not show any message about not being able to compile msgpack and msgpack itself will attempt to compile, and then will fallback to pure python without failing the install. If that changes your opinion on the matter at all.

StephanErb:
@ionrock I can absolutely relate to you, now wanting an additional dependency. I will reconsider doing the pull request directly in pip

ionrock:
@StephanErb @dstufft 

After some conversation in irc, I'd be open to merging this. @StephanErb do you mind taking a look at your patch to make sure things are still working?

This will also need some documentation updates to help communicate the change. 

Thanks!

dsully:
+1 - this would be great to have.

StephanErb:
I have rebased this pull request. No code modifications were necessary and all tests are still passing. From my perspective this would be ready to merge.

cachecontrol/serialize.py
```diff
@@ -3,23 +3,10 @@
...
-
-
-def _b64_encode_str(s):
-    return _b64_encode_bytes(s.encode("utf8"))
-
-
-def _b64_encode(s):
-    if isinstance(s, text_type):
-        return _b64_encode_str(s)
-    return _b64_encode_bytes(s)
```
ionrock: This should probably be left around for the time being as they are still used in the previous versioned loader. I would think that if a v3 cache item gets loaded we'll see an exception.

tests/test_serialization.py
```diff
@@ -105,33 +116,3 @@ def test_no_vary_header(self, url):
...
-        string.
-
-        This test ensures we recognize the unicode encoded string act
-        accordingly.
-        """
-        unicode_result = _b64_encode(self.unicode_string.encode('utf-8'))
-        assert _b64_decode_str(unicode_result) == self.unicode_string
-
-        bytes_result = _b64_encode(self.unicode_string)
-        assert _b64_decode_str(bytes_result) == self.unicode_string
```
ionrock: As per the comment regarding these functions, this test can stick around as well.

ionrock:
@StephanErb Just a couple small comments. Thanks for rebasing!

cachecontrol/serialize.py
StephanErb: I have only deleted the encode methods. The decode methods required for reading are still around. I therefore think that loading an old compressed json item is still working. There is also a test case for this. See my comment below.

I am happy re-add if the code. Just want to make sure we are on the same page :)

cachecontrol/serialize.py
```diff
@@ -3,23 +3,10 @@
...
+from .compat import HTTPResponse, pickle
 
 
 def _b64_decode_bytes(b):
```
StephanErb: Decode has not been removed to ensure backwards compatibility.

tests/test_serialization.py
```diff
@@ -30,17 +29,29 @@ def setup(self):
...
-    def test_read_version_two(self):
+    def test_read_version_v1(self):
         req = Mock()
         resp = self.serializer._loads_v1(req, pickle.dumps(self.response_data))
-        # We have to decode our urllib3 data back into a unicode
-        # string.
+        # We have to decode our urllib3 data back into a unicode string.
+        assert resp.data == 'Hello World'.encode('utf-8')
+
+    def test_read_version_v2(self):
```
StephanErb: This test ensures we can still read old data.

cachecontrol/serialize.py
StephanErb: Decode has not been removed to ensure backwards compatibility.

cachecontrol/serialize.py
ionrock: @StephanErb you were right! Thanks!

StephanErb:
I have rebased this branch. Is there anything else that need to be addressed?

ionrock:
Thanks @StephanErb for your patience with this patch!

dstufft:
💯

StephanErb:
@ionrock would it be possible to make a new release including this chance?

ionrock:
@StephanErb Yes! I just started at a new job and have been a bit swamped. I'll get it out today.

dsully:
@ionrock - any chance we can get a release soon? Thanks!

ionrock:
@dsully @StephanErb Just released 0.12.0. Let me know if you find any issues. Thanks!

PR #120, opened 2016-04-23
Include tests in release source tarball

EdwardBetts:
A test suite is a very useful thing, it would be great to include it in the pypi release tarball. That way people who download and build the release can run the tests to ensure the library is working.

Linux distributions, such as Debian, base their packages of Python modules on the pypi release. This means the test suite can be run when building a Debian package, it should catch mistakes in the packaging or errors in dependencies.

I've written some more on this topic on the Debian Python mailing list: https://lists.debian.org/debian-python/2016/04/msg00074.html

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include LICENSE.txt
+recursive-include tests *.py


ionrock:
@EdwardBetts Sorry for taking such a long time to provide some feedback on this. I understand the reasoning, but disagree that it is a good practice. I'd argue that as long as the building of packages is automated and audited (ie tests are run) then pulling from a repo and testing against a tag is a better course of action than expecting tests distributed with the package. Some code requires a great deal more testing and requires external elements to verify fully. For example, in the CacheControl suite, I start a cherrypy server to test some responses using a reasonably compatible HTTP server. In other projects I've used tooling such as docker containers for supporting services in a large test suite. I'd rather not guarantee that tests in the package can run when I distribute a package on pypi and prefer to request people trying to validate my packages take the same rigor by reading the README, understanding the dependencies and logic behind them in order to validate the package accordingly. 

All that said, I can certainly be convinced otherwise, but as it stands now, I'd prefer to keep them separate.

Issue #123, opened 2016-05-08
Should heuristic apply regardless of status code?


eads:
In my attempts to debug #122 (which turned out to be a quirk of my API provider), [I implemented a heuristic](https://github.com/newsdev/elex/blob/0121-response-caching/elex/cachecontrol_heuristics.py). The heuristic simply removes the max-age cache-control header. Voila, the etag-based caching worked, until the first cache hit (or so it seems). Once I got a cache hit, the max-age based debug messages were popping up again. 

What I realized is that the heuristic isn't applied to 304 responses (or 301s). My suspicion is that it should be -- [I moved the heuristic outside the conditional](https://github.com/eads/cachecontrol/commit/db28cafcdb1237309fd6b619c6c3319a014fd353) and now it seems to be working the way I would expect.

I need to dig around more -- this could be a case of getting age-related debug messages but the underlying mechanism still working correctly, but as far as I can tell now, not applying heuristics to 304s is a minor bug.

eads:
Sorry to bother you on this but do you have any plans to take a look at this and roll a release any time soon? No worries if you can't; if you're swamped I think what we'll do is roll a special release of our fork called elex-cachecontrol so that our package can use it as a dependency. In that case we'll put a big ass disclaimer on it and direct people to the main library for their needs, and then switch to the official version whenever you have a chance to take a look.

(fwiw we appreciate everything you've done with this and I'm not trying to be annoying)

ionrock:
@eads I took a look! Thanks for the PR! I'm happy to do a release as soon as we wrap up that PR.

ionrock:
Going ahead and closing this with 0.11.7 released. Thanks @eads!

PR #124, opened 2016-05-09
apply heuristics more aggresively

eads:


eads:
Whoops, I meant to put this in the PR description, but leaving it here now.

This is a possible solution to #123, though I think it needs some tests. 

I'm also curious if heuristics should apply to 301s as this PR would have it.

I'm also also curious if heuristics should apply even when method isn't GET, basically to give developers the maximum control over shaping the response that is sent for caching. That seems a little crazy but I figured it's at least worth asking.

My expectation with respect to heuristics was that they applied to any, or at least any cacheable, response.

ionrock:
@eads I think this makes tons of sense. Can you add a test?

eads:
I surely can. I should be able to get to this over the weekend. 

It seems like a good test would be one that replicates the bug when run against master and that passes with the PR. That's the direction I'll pursue unless you have other thoughts.

eads:
Hey @ionrock! I wrote that patch. It tests cleanly in py 2.x but not in 3.x -- but the error seems like a bug in the test itself, or possibly even just a spurious error. It seems the test I added is somehow causing the CherryPy server used by the testing rig to choke. I'm curious if you have any insight.

Otherwise I feel pretty confident in this -- tests fail when patch isn't applied and pass when it is applied.

eads:
Just a quick bump on this. I'm going to see if I can figure out what's up. We'd like to roll a final version of Elex by late July and this is one of the blockers. We can use a forked version as a dependency but that's no fun for anyone.

ionrock:
@eads Ok, I think the error was just travis killing the cherrypy process when that was called. I'm going to merge and see if I can ensure the test doesn't keep failing.

eads:
Huzzah! Any hunch as to why Travis was killing the cherrypy process? I couldn't find anything obvious in my tests that would cause a regression, but (far) stranger things have happened.

eads:
Bummer, it doesn't seem to have been resolved. I can take a look at this next week if that would help.

PR #129, opened 2016-10-07
Remove accidental lockfile ImportError message when using RedisCache

jezdez:


jezdez:
Since this requires a change of import statements I understand that this may not be the nicest thing to do for current users. Alternatively the message could be shown as a Python warning instead (which can be filtered out) or even only during the file cache backend initialization instead of during import time.

dsully:
+1 !

This fixes #130

ionrock:
@jezdez As this effects the file cache import can we consider a change that doesn't effect that import? I'm fine using warnings or something else (logging?) that can be turned off reasonably.

jezdez:
@ionrock Yeah, I understand where you come from and think the only way out of this is basically not have the warning or notice being shown during import time in the first place. Instead raise an ImportError when the `FileCache` class is instantiated as that makes `RedisCache` and `FileCache` behave similarly. `RedisCache` simply doesn't have that problem since it's passed a redis connection object as the only hard dependency, while `FileCache` does the leg work on its own.

Anyways, the PR reflects this, no API changes and less implicit import time side effects.

ionrock:
Thanks @jezdez! I appreciate you making that change!

jezdez:
@ionrock Happy to! Thanks for making the library :)

PR #140, opened 2017-02-20
Make wrapper take an optional CacheController option

kyrias:
This makes it simpler to override the CacheController to implement
customized behaviour.


elnuno:
This was added in #144.

PR #143, opened 2017-03-08
FileNotFoundError fallback for 2.X

grokcode:
This maintains compatibility for Python 2.X. Also see https://github.com/pypa/pip/issues/4322.


brad:
I'm seeing this too, haven't found any workaround :-(

PR #146, opened 2017-03-21
Correctly handle cached shared between Py2 and Py3

dstufft:
This issue a bit subtle, but essentially msgpack (the way it's being used here) has a Python3 like model with a strict split between bytes and text. This is a generally a good thing, HOWEVER it plays havoc when you're *creating* a cached file on Python 2.x and *consuming* it on Python 3.x. This is because the ``str`` type is both text and bytes on 2.x, so msgpack makes the assumption that ``str``/``bytes`` is binary data and ``unicode`` is textual data.

This all ends up meaning that serializing something like ``{"response": ...}`` on Python 2 ends up getting serialized the same as ``{b"response": ...}`` on Python 3. This causes things to blow up when attempting to load a cached entry serialized by Python 2 on Python 3, because it's is essentially doing this: ``{b"response": ...}[u"response"]``.

Thus what this patch does is force even ASCII text to be unicode when serializing out the data, which will cause msgpack to write them as the str type and all will be well. Because we can only infer (and we can't *know*) what the correct type was for already cached v3 entries, I went ahead and made v3 entries a miss and moved the cache entries onto v4.

This also includes a tweak to ``text_type`` to make it match the behavior of six better. Nothing appears to be currently using that so it should not break anything.

### DIFF
--- a/cachecontrol/compat.py
+++ b/cachecontrol/compat.py
@@ -24,6 +24,6 @@ except ImportError:
 
 # Replicate some six behaviour
 try:
-    text_type = (unicode,)
+    text_type = unicode
 except NameError:
-    text_type = (str,)
+    text_type = str
--- a/cachecontrol/serialize.py
+++ b/cachecontrol/serialize.py
@@ -6,7 +6,7 @@ import zlib
 import msgpack
 from requests.structures import CaseInsensitiveDict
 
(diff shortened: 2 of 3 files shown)


dstufft:
Thanks! A release would be great too :)

ionrock:
@dstufft 0.12.2 is released!

Issue #149, opened 2017-04-06
'Redis' object has no attribute 'disconnect'


kirberich:
Hi!

I just started using CacheControl with redis in a project, and while the caching itself works great, I'm running into problems when CacheControl tries to disconnect from redis, where it's calling a `disconnect` method on the redis object that doesn't exist.

I haven't been able to dig much deeper into this yet, some pointers would be much appreciated! I'll try to find out more about this tomorrow.

```python
.virtualenv/lib/python3.6/site-packages/requests/sessions.py:369: in __exit__
    self.close()
.virtualenv/lib/python3.6/site-packages/requests/sessions.py:690: in close
    v.close()
.virtualenv/lib/python3.6/site-packages/cachecontrol/adapter.py:129: in close
    self.cache.close()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <cachecontrol.caches.redis_cache.RedisCache object at 0x10f6f60f0>

    def close(self):
>       self.conn.disconnect()
E       AttributeError: 'Redis' object has no attribute 'disconnect'

.virtualenv/lib/python3.6/site-packages/cachecontrol/caches/redis_cache.py:41: AttributeError
```

ionrock:
@kirberich unfortunately, that backend isn't very well supported. It shouldn't be hard to fix if you're interested. I'm happy to merge a fix as well. That said, I'm unable to try and get a fix right now.

kirberich:
@ionrock thanks for the quick reply! 

I've sent a pull request that hopefully fixes this issue, it really just removes the connection closing from the RedisCache, which I think is the right thing to do with a redis cache that uses connection pooling.

Please let me know if you think this isn't the right fix!

Cheers
Rob

PR #150, opened 2017-04-07
Don't close redis connection when adapter gets closed

kirberich:
Hi!

As mentioned yesterday in #149, I found some problems with cachecontrol attempting to close the connection to redis when .close is called on the cache adapter, causing issues because the `Redis` object doesn't have a disconnect method.

I'm not an expert on redis, but as far as I can tell py-redis uses connection pooling by default, making it unnecessary to explicitly close connections (source: http://stackoverflow.com/questions/24875806/redis-in-python-how-do-you-close-the-connection)

I've coded up a quick fix that simply doesn't do anything when closing a redis cache, which I _think_ is the correct behaviour. There might be cases where a user manages to instantiate redis without a connection pool, in which case closing the connection might be necessary, but I don't even know how to do that, so I'm hoping we can shelve that potential problem for now.

For reference, here's a quick example of code I've been using that caused problems for me with the original behaviour:

```python
@contextmanager
def cache_session(session=None):
    """Context manager to easily apply caching to a session.

    An existing session can be passed in, otherwise
    a new Session is created.
    """

    session = session if session else Session()
    session.mount('http://', cache_adapter)
    session.mount('https://', cache_adapter)

    yield session
```


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.12.2
+current_version = 0.12.3
```
ionrock: Can we revert the bumpversion call please? I'll handle the version bump when a new version is released.

setup.py
```diff
@@ -18,7 +18,7 @@
     description='httplib2 caching for requests',
     long_description=long_description,
     install_requires=[
-        'requests',
+        'requests==2.12.4',
```
ionrock: We need to try and support as many versions of requests as possible, so this shouldn't be pinned. As an addon to requests, I consider it CacheControl's role to support the upstream requests dependency rather than define it.

kirberich:
Yeah sorry about those unrelated things, we were launching something today and I needed to make those changes to not break our app and forgot they were going to update the PR - I'll do some fiddling with branches to make sure the PR is separate from the stuff we need internally

ionrock:
@kirberich Totally understand! I'm not sure how you are releasing your code, but I'd expect the app that is using CacheControl to pin the requests version. A requirements.txt is a decent methodology for that aspect as it is separate from setuptools/distutils so you are coupled to the python build system for your deployment environment. For example, [pipfile](https://github.com/pypa/pipfile) is one example of how to help solve that sort of issue. 

Obviously you might already know this! But, it can't hurt to mention it :)

kirberich:
Yeah our issue was that pip and setuptools don't get along very well sometimes and there was a problem with cachecontrol pulling in the newest version of requests and other bits of our system specifying a concrete version in a requirements.txt, which resulted in a version conflict - it's a solvable problem but it came up right in the middle of a deployment so we just needed a quick fix, it wasn't meant to end up in the PR though ;)

I'm going to close this PR and create a new one from a different branch with just the fixing commit, sorry for the inconvenience!

PR #156, opened 2017-05-22
Modified alias of FileNotFoundError to OSError instead of IOError

vdaskalov:


vdaskalov:
Hi,
Can we have this prioritized somehow? It's kind of a blocker for me since it breaks the whole requests module. 

Looking forward for a release with the change :)

PR #163, opened 2017-07-10
Controller: Convert max-age to integer before comparsion

dkupka:
In parse_cache_control all keys and values are hadled as strings so the value
must be first converted to int otherwise the condition is never met.

### DIFF
--- a/cachecontrol/controller.py
+++ b/cachecontrol/controller.py
@@ -62,27 +62,51 @@ class CacheController(object):
         return cls._urlnorm(uri)
 
     def parse_cache_control(self, headers):
-        """
-        Parse the cache control headers returning a dictionary with values
-        for the different directives.
-        """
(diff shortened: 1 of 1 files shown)


ionrock:
@dkupka Great catch! Would you mind pulling this cast out to its own line in order to ensure that conversion doesn't throw any exceptions? As the header can be set to anything by the client, it is certainly possible a bad header would be sent that isn't an int.

A test that asserts a `max-age: foo` would be helpful as well to make sure we have known behavior when a bad header is sent.

dkupka:
@ionrock I wanted to keep the change minimal but OK :-) I've changed parse_cache_control method to parse, convert and validate cache-control directives.

ionrock:
Thanks @dkupka! Great refactoring!

PR #164, opened 2017-07-14
Use tox-travis

jayvdb:
Add Python 3.5 and 3.6 to tox, and run all tox testenvs in Travis CI.


jayvdb:
Intermittent problem in the tests; will likely fix itself with a restart of the job.  The same commit on my fork built successfully.

Issue #141, opened 2017-02-22
Proper `Cache-Control: private` support


kyrias:
Responses where `Cache-Control` is set to `private` may be cached but must not be cached in a shared cache.  I have a use of cachecontrol for an API client I'm writing, but it has support for multiple users, so I need to implement user-specific caching support soon.

My idea for how to implement it is to have a parameter you pass to `CacheControl` with a header name.  Then when a response has `Cache-Control: private` set it would try to fetch a header with that key from the request/response and then then prepend a hash of that to the cache key.

I currently have it implemented locally for the redis backend, though it's currently very implementation specific, only checking for `Authorization`, and only for the redis backend.

Any thoughts/opinion on this idea?

ionrock:
That is a tough one! I'm curious about what qualifies as "shared". For example, if I were using the file cache, would using a different directory be enough to be considered private? I suppose the same goes for the redis cache where private and public items might be cached. 

As for an implementation, I *think* what I would do is define something like a `namespace` function that is passed to the `CacheControl` wrapper. If defined, the function would return a prefix key that would be prepended when storing it in the cache. 

That sounds pretty similar to what your doing with the exception being that a function is passed in that allows accessing anything in the request to be used when constructing the private namespace. 

I hope that helps a bit! It sounds like an interesting problem and I'm interested to see how you solve it!

kyrias:
Essentially it shouldn't be shared as in it shouldn't be possible for the result to be returned for other users.

I do like the idea of a `namespace` function though, hm.  I'll try it out, thanks!

jowrjowr:
This issue just dinged me pretty hard.

Looking at the standard for this:

https://tools.ietf.org/html/rfc7234#page-21

```   The "private" response directive indicates that the response message
   is intended for a single user and MUST NOT be stored by a shared
   cache.  A private cache MAY store the response and reuse it for later
   requests, even if the response would normally be non-cacheable.```

This is something cachecontrol should not be doing at all, unless specifically enabled in a heuristic which you already have sufficient support for. 

I was going mental trying to figure out how something with Cache-Control: private was getting cached and then I ripped apart the cachecontrol code proving it. 

Pretty please don't break standards like that :(

ionrock:
[...]

PR #167, opened 2017-08-07
Allow defining a cache as "shared"

ionrock:
In the spec (https://tools.ietf.org/html/rfc7234#section-5.2.2.6) it talks
about the `private` directive and that a shared cache MUST NOT store
the response.

This starts by defining the cache as being "shared" and testing for
the "shared" attribute on the cache.

Fixes: #141


Flameeyes:
FTR, I have something along these lines in #225 as well.

PR #169, opened 2017-08-23
zlib compression of cache objects

jowrjowr:
Does exactly what it says on the tin.

Before I caught onto the madness with redis, my redis requests cache store was 3 gigabytes. Granted it took awhile to do that, but I can't be the only one. Using zlib compression will save a reasonable amount of memory for the memory-based caches cachecontrol can use.

Ought to merge clean with the redis PR. This was written separately. 

Passed tests on my machine.

### DIFF
--- a/cachecontrol/cache.py
+++ b/cachecontrol/cache.py
@@ -3,7 +3,7 @@ The cache object API for implementing caches. The default is a thread
 safe in-memory dictionary.
 """
 from threading import Lock
-
+import zlib
 
 class BaseCache(object):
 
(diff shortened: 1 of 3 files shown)


StephanErb:
A universal compression will cause significant slowdowns when we are caching already compressed files such as images, tarballs, or zip files. I therefore believe that the decision to enable compression needs to be a bit more fine grained.

jowrjowr:
That is true. The immediate problem I see is that if it is written that
way, there's no obvious way for the get() call to realize it is
uncompressed.

Maybe a secondary key to indicate compression? That feels convoluted.

I'm not clear that zlib compression of already-compressed things would add
enough overhead for it to justify special casing.

ionrock:
This feels orthogonal to me. I see how it could be helpful (caching large objects), but at the same time, I'd also advocate a user create a specialized cache and not rely on the default redis cache, which was admittedly added as an example more so than something production ready.

The idea might be helpful over time, but I think it needs more thought to be done correctly. Going  to close this one, but @jowrjowr, if you do make a specialized cache that supports compression, feel free to update the docs with links and a description. I'm always happy to promote helpful cache implementations, especially because I've tried to avoid adding them to CacheControl proper.

PR #171, opened 2017-09-20
make CacheControlAdapter picklable

tastuteche:
make CacheControlAdapter picklable,so ready for asyncio->run_in_executor + ProcessPoolExecutor
(1)when I run https://github.com/tastuteche/tell-me-more/blob/master/tell_me_more/github_title.py under python3.5
got error:
concurrent.futures.process._RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/r/a/p/usr/lib64/python3.5/concurrent/futures/process.py", line 175, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/r/a/p/usr/lib64/python3.5/site-packages/requests/sessions.py", line 515, in get
    return self.request('GET', url, **kwargs)
  File "/r/a/p/usr/lib64/python3.5/site-packages/requests/sessions.py", line 502, in request
    resp = self.send(prep, **send_kwargs)
  File "/r/a/p/usr/lib64/python3.5/site-packages/requests/sessions.py", line 612, in send
    r = adapter.send(request, **kwargs)
  File "/r/a/p/usr/lib64/python3.5/site-packages/cachecontrol/adapter.py", line 39, in send
    cacheable = cacheable_methods or self.cacheable_methods
AttributeError: 'CacheControlAdapter' object has no attribute 'cacheable_methods'
"""
(2)found a fix at stackoverflow
The ProcessPoolExecutor class is an Executor subclass that uses a pool of processes to execute calls asynchronously. ProcessPoolExecutor uses the multiprocessing module, which allows it to side-step the Global Interpreter Lock but also means that only picklable objects can be executed and returned.
https://stackoverflow.com/questions/17419879/why-i-cannot-use-python-module-concurrent-futures-in-class-method

### DIFF
--- a/cachecontrol/adapter.py
+++ b/cachecontrol/adapter.py
@@ -31,6 +31,21 @@ class CacheControlAdapter(HTTPAdapter):
             serializer=serializer,
         )
 
+    def __getstate__(self):
+        state = super(CacheControlAdapter, self).__getstate__()
+        state['cache'] = self.cache
+        state['heuristic'] = self.heuristic
+        state['cacheable_methods'] = self.cacheable_methods
+        state['controller'] = self.controller
+        return state
+
+    def __setstate__(self, state):
+        super(CacheControlAdapter, self).__setstate__(state)
+        self.cache = state['cache']
+        self.heuristic = state['heuristic']
+        self.cacheable_methods = state['cacheable_methods']
+        self.controller = state['controller']
+
     def send(self, request, cacheable_methods=None, **kw):
         """
         Send a request. Use the request information to see if it


ionrock:
@tastuteche Would you mind clarifying the use case a little more? I understand you're trying to use something that is trying to pickle the session, which fails due to the adapter not being pickleable, but I'd like to get a little more insight into why you're trying to pickle a requests session?

I ask b/c the intent for sharing a cache is to use the cache storage instead. For example, rather than passing the requests session and adapter, you'd pass along the cache storage info and create a new session and cache control adapter in the child worker (ie the executor you mentioned). 

I also will mention that when I've tried to pickle complex objects, it has typically ended up being a bad experience with confusing bugs. By doing a little work to pass data, most of the concurrency bugs go away and things work as expected. I mention my experiences in Python here in case it might be something you're experiencing as well. Other languages provide better concurrency systems where your design (as I understand it) makes perfect sense.

tastuteche:
a legacy system,having 2 parts
(1)a map reduce framework , doing io and cpu scheduling by ProcessPoolExecutor and ThreadPoolExecutor.
(2)the wokers,basically doing io(requests + cachecontrol) ->nlp processing stuff
part1 would be replaced by asnycio, a elegant way  for io scheduling(waiting for workers doing cpu-bound job).
but part2 is not asyncio ready. part2 should be splited into another two parts:
(2.1) http + web cache
(2.2) doing nlp processing stuff on web content
here is some example code
--------------
common part of legacy code1 and code2:
session = requests.Session()
sess = CacheControl(session,
                    cache=FileCache('.webcache', forever=True),
                    heuristic=ExpiresAfter(days=1))
--------------                    
legacy code1:
def get_title(url):
    soup = bs4.BeautifulSoup(sess.get(url).text, "lxml")
    ...
    
with futures.ProcessPoolExecutor(max_workers=4) as executor:
    for url, text in executor.map(get_title, urls):
        ...
----------------
code2:
p = ProcessPoolExecutor(2) 
async def get_title(url):
    future1 = loop.run_in_executor(p, sess.get, url)
    response1 = await future1
    soup = bs4.BeautifulSoup(response1.text, "lxml")
    ...

async def main():
    futures = map(lambda url: get_title(url), urls)
    all_data = await asyncio.gather(*futures)
        ...
-------------
code1 works well when CacheControlAdapter not pickleable or pickleable
code2 only works when CacheControlAdapter pickleable

code2 should be replaced by aiohttp + async web cache,if async web cache implements are available.
but now,i would like to make CacheControlAdapter pickleable,as legacy code works well with requests + cachecontrol,
and asyncio can interact with legacy code by using asyncio.get_event_loop().run_in_executor.
the problem ,it seems that run_in_executor is more strict about pickleable then legacy ProcessPoolExecutor,as code1 works well and code2 not working when  CacheControlAdapter not pickleable.

ionrock:
I'm going to close this one. It seems better to refactor the `get_title` code to create a session when necessary such that you don't have to pickle and pass the cachecontrol object around. I can see why you'd want do that since you can keep a session open, but if you already have to pickle it, you likely don't get to keep the threadpool the requests session provides.

PR #172, opened 2017-11-17
Add named arguments to setex redis cache method

krzysztofzuraw:
It fixes `value is not an integer or out of range` when using `set` for `RedisCache` with expire time


jaap3:
This seems related to #168, where the same issue was found

ionrock:
This was already merged it seems.

PR #177, opened 2018-02-01
Prepended cachecontrol to adapter in usage example

aJetHorn:
Since only cachecontrol was imported, it must also be used in the call to CacheControlAdapter()

This is a trivial change but it's also my first pull request ever -- looking forward to contributing in a more substantial way.


ionrock:
Thanks @aJetHorn!

PR #178, opened 2018-02-16
Fix exception raising in BaseCache

jwilk:
`NotImplemented` is a sentinel object, not an exception type.
It was mistakenly used instead of `NotImplementedError`.


ionrock:
@jwilk Good catch! Thanks!

PR #182, opened 2018-03-26
Fixed #179: Catch potential FileNotFoundError

### LINKED ISSUE #179: FileCache: No such file or directory <file>
Hello,

I've just started using cachecontrol with FileCache so I apologize if I'm mistaken somewhere.

When using it with the standard, in memory, cache, it works fine, but when switching to FileCache I get a few errors such as:  [Errno 2] No such file or directory: '.webcache/a/9/7/a/2/a97a2bcda25b30914b0e070e50724edfb85cc70f45ece690e0fbfb98'

The error does not happen that often, maybe a few times out of thousands of calls, and it seems on the same files so there may be a pattern there.

After my script crashes I can verify that the file is there and not empty.

My script has 8 threads calling get_page(url) in parallel, potentially asking for the same url at the same time.

The code I use for requests/cachecontrol:

    from cachecontrol import CacheControl
    from cachecontrol.caches.file_cache import FileCache
    import requests

    session = requests.Session()
    session.headers = {'User-Agent': 'Mozilla/5.0'}
    session.cookies = requests.cookies.RequestsCookieJar()
    session.allow_redirects = True
    cached_session = CacheControl(session,
                                  cache=FileCache('.webcache', forever=True))
    
    def get_page(url):
        try:
            return cached_session.get(url)
        except requests.exceptions.TooManyRedirects:
            return None

Is it related to the TooManyRedirects, in that requests would crash, and maybe cachecontrol would not have the file ready even though it thought it would?

Switching to one thread seems to alleviate the problem, though I have not yet ran my script like this fully to really know.

Python: 3.6.4
requests: 2.18.4
cachecontrol: 0.12.4
lockfile: 0.12.2

running 'pip install cachecontrol[filecache]' added msg-pack-python-0.5.5 but the behavior did not change.

I am happy to provide any information needed.

Thank you!

jaap3:
When multiple processes share the same file cache, a file can potentially be deleted between the check for existence (`if not os.path.exists(name)`) and the actual read operation.

### DIFF
--- a/cachecontrol/caches/file_cache.py
+++ b/cachecontrol/caches/file_cache.py
@@ -9,7 +9,7 @@ try:
     FileNotFoundError
 except NameError:
     # py2.X
-    FileNotFoundError = OSError
+    FileNotFoundError = (IOError, OSError)
 
 
 def _secure_open_write(filename, fmode):
--- a/cachecontrol/caches/file_cache.py
+++ b/cachecontrol/caches/file_cache.py
@@ -95,11 +95,11 @@ class FileCache(BaseCache):
 
     def get(self, key):
         name = self._fn(key)
-        if not os.path.exists(name):
+        try:
+            with open(name, 'rb') as fh:
+                return fh.read()
+        except FileNotFoundError:
             return None
-
-        with open(name, 'rb') as fh:
-            return fh.read()
 
     def set(self, key, value):
         name = self._fn(key)


jaap3:
The fail on Python 2.7 is because FileNotFoundError is an alias of OSError instead of IOError. This alias was changed from IOError to OSError in 5ed4b96, I'm not sure why.

jaap3:
@ionrock is there anything I can do to get this PR merged? I'm seeing this happen (infrequently) in a production application and am eager to get this fixed.

ionrock:
@jaap3 Nope! That looks great. Thanks for your patience.

PR #183, opened 2018-04-08
Fix behaviour of Cache-Control: no-store on responses

sk1p:
According to RFC 7234, section 5.2.2.3: Response Cache-Control
Directives, no-store:

> The "no-store" response directive indicates that a cache MUST NOT
> store any part of either the immediate request or response.  This
> directive applies to both private and shared caches.  "MUST NOT
> store" in this context means that the cache MUST NOT intentionally
> store the information in non-volatile storage, and MUST make a
> best-effort attempt to remove the information from volatile storage
> as promptly as possible after forwarding it.[...]

https://tools.ietf.org/html/rfc7234#section-5.2.2.3

The second part ("[...] MUST make a best-effort attempt to
remove the information [...]") was already handled, but if any other
conditions for caching responses became true, for example presence of
Etags, permanent redirect etc. the response was cached even with
``Cache-Control: no-store``.

This behaviour is fixed by just returning from ``cache_response`` if the ``no_store`` flag is set,
potentially after having purged the cache entry.

### DIFF
--- a/tests/test_cache_control.py
+++ b/tests/test_cache_control.py
@@ -115,21 +115,11 @@ class TestCacheControllerResponse(object):
         cc.cache_response(self.req(), resp)
         assert not cc.cache.get(cache_url)
 
-    def test_cache_response_no_store_with_etag(self):
-        resp = Mock()
-        cache = DictCache({self.url: resp})
-        cc = CacheController(cache)
-
-        cache_url = cc.cache_url(self.url)
-
+    def test_cache_response_no_store_with_etag(self, cc):
         resp = self.resp({'cache-control': 'no-store', 'ETag': 'jfd9094r808'})
-        assert cc.cache.get(cache_url)
-
-        # skip serializer as it can't handle mocks
-        cc.serializer = Mock()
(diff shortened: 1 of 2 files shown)


sk1p:
The Travis failure appears to be a flaky test - I was able to reproduce it locally, but only once. Has happened before: https://travis-ci.org/ionrock/cachecontrol/jobs/191119174

tests/test_cache_control.py
```diff
@@ -115,6 +115,22 @@ def test_cache_response_no_store(self):
         cc.cache_response(self.req(), resp)
         assert not cc.cache.get(cache_url)
 
+    def test_cache_response_no_store_with_etag(self):
```
jaap3: I think this test can be reduced to this:

    def test_cache_response_no_store_with_etag(self, cc):
        resp = self.resp({'cache-control': 'no-store', 'ETag': 'jfd9094r808'})
        cc.cache_response(self.req(), resp)

        assert not cc.cache.set.called

ionrock:
Thanks @sk1p !

PR #190, opened 2018-06-05
Convert headers to text when serializing

pradyunsg:
Headers are also text and on Py2/Py3 should be treated as unicode/str instead of
the str/bytes.

Seems to have been missed in #146.

/cc @dstufft @ionrock

### DIFF
--- a/cachecontrol/serialize.py
+++ b/cachecontrol/serialize.py
@@ -64,7 +64,7 @@ class Serializer(object):
         if u"vary" in response_headers:
             varied_headers = response_headers[u'vary'].split(',')
             for header in varied_headers:
-                header = header.strip()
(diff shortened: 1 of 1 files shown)


ssbarnea:
I hope to see a release too! I have nightmares scrolling the "Cache entry deserialization failed, entry ignored" logs.

ssbarnea:
@dstufft @ionrock Can you please made a new release for this so we can fix pip and other packages affected by it?

dstufft:
I can't make a release, I'm just a consumer of cachecontrol :)

ssbarnea:
Indeed, not sure if this is the right approach but I made https://github.com/ionrock/cachecontrol/pull/192 -- probably not enough to add the tag.

PR #197, opened 2018-11-23
Make v4 cache compatible with msgpack 1.0 in the future

yan12125:
The `encoding` parameter in `msgpack.loads` is deprecated and will be
removed in msgpack 1.0 [1].

Note that the `raw` parameter was called `raw_as_bytes` before 0.5.2
[2]. I add the version constraint to avoid API incompatibilities.

[1] https://github.com/msgpack/msgpack-python/issues/191
[2] https://github.com/msgpack/msgpack-python/commit/5569a4efcdc913d343eaff4e55c9b19fafde4268

### DIFF
--- a/cachecontrol/serialize.py
+++ b/cachecontrol/serialize.py
@@ -181,7 +181,7 @@ class Serializer(object):
 
     def _loads_v4(self, request, data):
         try:
-            cached = msgpack.loads(data, encoding="utf-8")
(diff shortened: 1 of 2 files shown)


yan12125:
Tests pass with Python 3.4 on my machine. No idea why they fail on Travis...

ionrock:
Thanks @yan12125 !

PR #201, opened 2019-01-25
Fixed test setup for pytest >= 4.1

TheJokr:
The `pytest_namespace` hook was removed in pytest 4.0 according to [their website](https://docs.pytest.org/en/latest/deprecations.html#pytest-namespace), but appears to work fine until version 4.1. Using that version yields a `pluggy.manager.PluginValidationError: unknown hook 'pytest_namespace'`.
This hook is used in `tests/conftest.py` to add `cherrypy.server` to the `pytest` namespace. Since the server instance is available through the `cherrypy` module globally anyway, I replaced the only reference to it (in the server fixture in the same file) by a reference to `cherrypy.server` and swapped `pytest_namespace` for `pytest_configure` for the initial setup of the server. This allows the tests to run as usual.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -107,7 +107,7 @@ class SimpleApp(object):
 
 @pytest.fixture(scope="session")
 def server():
-    return pytest.server
+    return cherrypy.server
 
 
 @pytest.fixture()
@@ -124,7 +124,7 @@ def get_free_port():
     return ip, port
 
 
-def pytest_namespace():
+def pytest_configure(config):
     cherrypy.tree.graft(SimpleApp(), "/")
 
     ip, port = get_free_port()
@@ -136,7 +136,6 @@ def pytest_namespace():
     logger.removeHandler(logger.handlers[0])
 
     cherrypy.server.start()
-    return {"server": cherrypy.server}
 
 
 def pytest_unconfigure(config):


TheJokr:
There seems to be a bug related to the mock server's `update_etag` route when reusing connections in some Python versions. The exception thrown in the 3.4 build happens when the connection used before that HTTP call is not reset (see for example [the build logs here](https://travis-ci.org/TheJokr/cachecontrol/builds/484602168)). I have only ever observed it on Python 3.4 and 3.5 builds, which could indicate that it's related to the [changes in `http.client` in Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#http-client). It is the same error as observed in #197.

TheJokr:
I think I found the bug causing these errors. A 304 response to a request is not allowed to have a message body (per [RFC 7232 section 4.1](https://tools.ietf.org/html/rfc7232#section-4.1)), but the test server [returned one anyway](https://github.com/ionrock/cachecontrol/blob/0fedbba7b740f0d03596adcf0fd88cdb2f1e653a/tests/conftest.py#L63:L69). If the next request was then sent fast enough, urllib3 reused the connection. But since the data was not read ([length is set to 0](https://github.com/urllib3/urllib3/blob/a90d3f3b8410c6d98fa21722d3ac66214819a702/src/urllib3/response.py#L298:L299)), it will be read instead of the status line on the following request, leading to the exception. I fixed it in the second commit.

yan12125:
> I think I found the bug causing these errors.

Awesome! You fixed an issue dated back to 2017 [1]. Hopefully @ionrock will review this patch soon.

[1] https://travis-ci.org/ionrock/cachecontrol/jobs/191119174

PR #202, opened 2019-01-25
Removed unnecessary caching of Vary: * responses

TheJokr:
When using cachecontrol in one of my projects I noticed a recurring warning about failed cache deserializations in my logs. After some investigation, I could pin the problem down to requests to https://updates.eveonline.com/rss/. As it turns out, the server sends a `Vary: *` header with its responses to that URL, which are cached by cachecontrol due to the also present `Cache-Control: public, max-age=60` header.
When later looking up a cached response to that URL, the cache returns None due to the `Vary: *` header ([see here](https://github.com/ionrock/cachecontrol/blob/0fedbba7b740f0d03596adcf0fd88cdb2f1e653a/cachecontrol/serialize.py#L108:L111)), which leads to the warning being emitted. Since these responses can never be served from the cache, storing them can be avoided altogether. This gets rid of the warning.

These changes depend on the changes from #201 for the tests to run correctly.


TheJokr:
I rebased this onto the fix for the bug discovered in #201.

ionrock:
Thanks!

PR #206, opened 2019-05-21
Added support for expiry on the Redis backend.

roblinton:
The CacheController does not pass an expiry key to supported backends, aka. Redis. This commit adds support for deriving a TTL from the response headers and passing that to the cache backend allowing the cache to expire old caches automatically.

When using this library in an environment with lots of traffic and rapidly expiring caches we noticed that the Redis backend never received an expiry argument.  We added this before noticing an existing pull request for the same functionality.  That pull request appears to be fairly old and no longer merges so I've updated our changes to incorporate some of the notes on the old one.

Some notes:
 - I changed the argument to RedisCache.set() from a date to an int and put the logic for determining the expiry time in CacheController.get_cache_expiry()
 - I changed the Redis method from setex(name, expiry, value) to set(name, value, ex=expiry) to avoid the whole argument order affair.
 - I added CacheController.CACHE_TTL_MAX to avoid storing keys forever in Redis and allow users to enforce a maximum cache duration (defaults to 2 weeks).  This is potentially outside the scope of this feature, if it creates problems let's drop it.

### DIFF
--- a/cachecontrol/caches/redis_cache.py
+++ b/cachecontrol/caches/redis_cache.py
@@ -13,7 +13,7 @@ class RedisCache(BaseCache):
         return self.conn.get(key)
 
     def set(self, key, value, expires=None):
-        self.conn.set(key, value, ex=expires)
+        self.conn.set(key, value, ex=expires or None)
 
     def delete(self, key):
         self.conn.delete(key)
--- a/cachecontrol/caches/redis_cache.py
+++ b/cachecontrol/caches/redis_cache.py
@@ -13,11 +13,7 @@ class RedisCache(BaseCache):
         return self.conn.get(key)
 
     def set(self, key, value, expires=None):
-        if not expires:
-            self.conn.set(key, value)
-        else:
-            expires = expires - datetime.utcnow()
-            self.conn.setex(key, int(expires.total_seconds()), value)
+        self.conn.set(key, value, ex=expires)
 
     def delete(self, key):
         self.conn.delete(key)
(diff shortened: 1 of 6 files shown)


cachecontrol/caches/redis_cache.py
```diff
@@ -13,11 +13,7 @@ def get(self, key):
         return self.conn.get(key)
 
     def set(self, key, value, expires=None):
-        if not expires:
-            self.conn.set(key, value)
-        else:
-            expires = expires - datetime.utcnow()
-            self.conn.setex(key, int(expires.total_seconds()), value)
+        self.conn.set(key, value, ex=expires)
```
bugflush: if  expires is 0: redis.exceptions.ResponseError: invalid expire time in set

roblinton: First off, my apologies for not having noticed this response until now!

Would it be better to make sure [get_cache_expiry()](https://github.com/ionrock/cachecontrol/pull/206/files/c9c98e26ac290fd31f4cbae2cf9ffc1e1bcf165e#diff-3a5ce26da48f679d11ee0eeb9d57ee58R347) never returns 0 (returning None instead in that case) or do validation in place (something like: ex=expires or None)?  Or both?

roblinton: I've done just the latter - coerce a 0 into a None in set().

We don't win much in putting more logic in get_cache_expiry() unless this logic becomes necessary in other parts of the library.  If a really wrong value is passed, it feels correct to raise an exception.

uraniun:
Hello @ionrock, any chance of this getting merged?

roblinton:
Implemented with https://github.com/ionrock/cachecontrol/pull/233

PR #210, opened 2019-08-30
uwsgi cache backend

etene:
Hello,

This PR adds support for [uwsgi's cache](https://uwsgi-docs.readthedocs.io/en/latest/PythonModule.html#cache-functions) as a backend, + a unit test.

Of course this backend only works in a uwsgi context; importing it outside of uwsgi will work, but instantiating `UWSGICache()` won't.

I also added a section to the relevant part of the documentation.

### DIFF
--- a/docs/storage.rst
+++ b/docs/storage.rst
@@ -101,6 +101,7 @@ Third-Party Cache Providers
 ===========================
 
 * cachecontrol-django_ uses Django's caching mechanism.
+* cachecontrol-uwsgi_ uses uWSGI's caching framework.
 
 
 
@@ -109,3 +110,4 @@ Third-Party Cache Providers
 .. _requests 2.1: http://docs.python-requests.org/en/latest/community/updates/#id2
 .. _redis: https://github.com/andymccurdy/redis-py
 .. _cachecontrol-django: https://github.com/glassesdirect/cachecontrol-django
+.. _cachecontrol-uwsgi: https://github.com/etene/cachecontrol-uwsgi


etene:
I forgot to mention that I also added a `clear()` method to `DictCache`, because I use it as a drop-in replacement when running my code outside of uwsgi, so I need it to have the same API as `UWSGICache`.

cachecontrol/cache.py
```diff
@@ -37,3 +37,7 @@ def delete(self, key):
         with self.lock:
             if key in self.data:
                 self.data.pop(key)
+
+    def clear(self):
```
ionrock: I understand that needed to use this method as a user of this backend? I'd go ahead and just add it to the cache back end and advertise its availability for users. I think you also said you fall back to the DictCache, in which case, you could also adjust the import error handling to return a compatible DictCache based cache if necessary.

cachecontrol/caches/uwsgi_cache.py
```diff
@@ -0,0 +1,40 @@
...
+from cachecontrol.cache import BaseCache
+from datetime import datetime
+
+class UWSGICacheError(Exception):
+    pass
+
+class UWSGICache(BaseCache):
+    def __init__(self, name=None):
+        if not uwsgi:
+            raise UWSGICacheError("uwsgi module is not available")
```
ionrock: To clarify on my previous comment in cache.py regarding the `clear()`, I think you could adjust this to do something like:

```
try:
    import uwsgi
    # Add your existing code
except ImportError:
   # Log a warning saying you're using a dict cache
   class UWSGICache(DictCache):
        def clear():
            pass
```
This way you're avoiding the exception when creating the object.

The reason you want to avoid errors from __init__ methods is that it makes it much harder to use the class. Every instantiation needs to be wrapped in a try/except block, which will change the scope of the variable making it hard to use. For example, often times when testing something, I'll instantiate some object purposefully skipping a critical attribute. The reason being is adding the extra data adds confusing noise, when my goal is to mock a specific method or use a method that doesn't require the attribute. 

I hope that is at least partially helpful describing why raising this error here is problematic.

docs/storage.rst
```diff
@@ -97,6 +97,31 @@ Here is an example using a `RedisCache`: ::
 This is primarily a proof of concept, so please file bugs if there is
 a better method for utilizing redis as a cache.
 
+UWSGICache
```
ionrock: My hope is that you can host this backend in its own repo, in which case, you can move the summary to the `Third-Party Cache Providers` and provide a link to help folks find it!

cachecontrol/caches/uwsgi_cache.py
etene: I don't really like the idea of the uwsgi cache still "working" when uwsgi is not present, and a simple warning might go unheeded in a production environment. But I get your point and it's still a better compromise than raising an exception in `__init__`, which I wasn't really comfortable doing anyway. So I'll make the change. Thanks !

etene:
No problem and thanks for the advice, it's appreciated. I'll move this to my repo & edit the PR to just point to it in the docs.

I'm pretty new to this kind of packaging so I'll have to look at how other external cache backends are implemented, though.

docs/storage.rst
etene: Will do !

cachecontrol/cache.py
etene: So you're saying you'd rather not have a `clear()` method in `DictCache`, and that I should just subclass it & add the method if needed ? That's ok with me, I'm just not 100% sure I understood your comment, English is not my 1st language.

etene:
Hello, 
I've moved the feature to its own repo (https://github.com/etene/cachecontrol-uwsgi), and rewritten the PR to mention it in the docs.

etene:
Hello @ionrock, any chance of this getting merged  ? The changes are pretty trivial now

PR #225, opened 2020-04-20
[RFC] (Tracking only) Reduce interface of cachecontrol, and make more reliable.

Flameeyes:
This includes the heavy-lifting changes that I don't feel are ready for actual review yet.


frostming:
Please adapt the code to the latest main branch and send another PR if necessary.

PR #233, opened 2020-09-29
add expires parameter to base class

nholmes3:
This PR is based on https://github.com/ionrock/cachecontrol/pull/168 but I have removed all the Redis related code as it is not needed for my use case. 

I have some other cache types based on BaseCache and would like to be able to pass through an expires parameter so that I can set a custom TTL in our backend cache.

### DIFF
--- a/cachecontrol/cache.py
+++ b/cachecontrol/cache.py
@@ -14,7 +14,7 @@ class BaseCache(object):
     def get(self, key):
         raise NotImplementedError()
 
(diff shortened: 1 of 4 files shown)


nholmes3:
Hello, @ionrock. Do you guys have a timeline for the when the next version will be released? It would be great to have this functionality available now that it is merged. If there is anything I can do to assist in the process, please let me know. I would be happy to help.

PR #234, opened 2020-10-06
Cache from status code 302

eamanu:
On issue https://github.com/pypa/pip/issues/8643 log we can see
that github response with a 302 status code [0]. From my point
of view status code 302 is similary to 301 that already is
accepted by cachecontrol, so shouldn't represent a problem
support 302 status code to cache.

[0] https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302

### DIFF
--- a/cachecontrol/adapter.py
+++ b/cachecontrol/adapter.py
@@ -96,7 +96,7 @@ class CacheControlAdapter(HTTPAdapter):
 
                 response = cached_response
 
-            # We always cache the 301 responses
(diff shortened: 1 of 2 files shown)


ionrock:
Looking at the [RFC](https://tools.ietf.org/html/rfc7231#section-6.4.3) the intent is that it is a temporary redirect that is expected to change over time, so I think caching it doesn't seem like the right thing to do by default. 

I'm going to go ahead and close this.

Issue #180, opened 2018-03-04
Catch MemoryErrors and give up instead of letting them propogate up


pradyunsg:
FYI - pypa/pip#2984

It might still exist with the current development version. =)

woodruffw:
Yep, resolved with #145.

Issue #145, opened 2017-03-15
Large responses cause increased memory usage.


dsully:
When downloading large files, memory usage is not constant when using CacheControl.

I believe this is due to the FileWrapper that buffers the response in memory.

If using requests directly:

```
import shutil
import requests

response = requests.get(url, stream=True)
with open('/var/tmp/out.bin') as fh:
    shutil.copyfileobj(response.raw, fh)
```

Yields constant memory usage. If you throw CacheControl into the mix, memory shoots up based on the size of the downloaded object.

dsully:
Digging more - the call to:

```self.serializer.dumps(request, response, body=body)```

In ```def cache_response(self, request, response, body=None):``` is looking like the culprit.

**Comment by elnuno** [comment]
ISTM that there might be an extra copy of the body attached to the response making things worse (`serialize.py:38`). I'll check for any low hanging fruits and try to trace it to see whether we could somehow use streaming, but I expect it not to be practical.

**Comment by elnuno** [comment]
Turns out there is a simple issue that might gives us most of the possible savings on a simple `CacheControl(requests.Session())` case: we tee the response `.read` into a buffer we never bother to close. Closing it makes things better (on Windows), in that total memory doesn't rise as much as in original code. PR incoming!

For patched code and bare requests I get:
> Using **32 MB** on program end.
> Mean memory use: **32 MB**

For original code I get:
> Using **106 MB** on program end.
> Mean memory use: **199 MB**

FWIW, I see more fluctuations in memory use even in the patched version (as opposed to raw requests) when watching the process in task manager. But it does't climb nearly as high as the unpatched version. Testing on *nix very welcome :)

Can you show how exactly you "throw CacheControl into the mix"? I'm using `sess = cachecontrol.CacheControl(requests.Session())`, but maybe using other caches makes things worse.

Thanks for reporting this :)

I'm using a test file like this:
```python
out = open('bigdata.bin', 'w', 'replace')
data = ''.join(chr(x) for x in range(37, 1064)) + '\n'
for i in range(70):
     for j in range(1024):
         out.write(data)
         out.write(str(i + j) + '\n' * 20)
out.close()
```

Then testing it like this:
```import os
import logging
import shutil

import requests
import cachecontrol
import psutil

# logging.basicConfig(level=logging.DEBUG)

us = psutil.Process(os.getpid())
MB = 1024 * 1024
N = 15

sess = cachecontrol.CacheControl(requests.Session())

total = 0
[...]

PR #240, opened 2021-02-16
Redisign the cahce API

hexagonrecursion:
The current cache API is:
```py
class Cache:
    def get(self, key):
        pass

        pass

    def delete(self, key):
        pass

    def close(self):
        pass
```

This makes fixing #145, #180 and #238 difficult because you have to store the entire response body somewhere before you can save it with `Cache.set()`.

Currently CacheControl buffers the response body in memory which is just **wrong** - thou shalt not assume that there are no files large enough to not fit in RAM. This is the cause of #145 and #180. Additionally this makes CacheControl completely unusable with response bodies larder than `(2^32)-1 bytes` no matter how much RAM you have due to a design limitation of the msgpack format #238.

I have considered a less invasive approach of buffering the response body in a temporary file, but there are at least two problems with this:
1. On Linux `/tmp` is often a `tmpfs` - a filesystem backed by virtual memory. It can swap out, but `swap` is not always enabled. `tmpfs` has a configurable maximum size limit. `swap` also often has a size limit. Basically: this would defeat the point - we would not be able to reliably cache a 100GB response even if there is enough storage for the cache.
2. Unoptimal IO: to cache a 10GB response we would write 10GB, read 10GB and write 10GB again instead of writing 10GB once.


hexagonrecursion:
Fixing #238 would also require a new serialization format because cc=4 stores both the body and the header in a single `msgpack` object.

ionrock:
Hey! Thanks for looking at this. I know this limitation has been considered before, but the discrepancies between different operating systems and potential use cases made it challenging to find a valid fix. The result was we did the simplest thing! One idea I had when looking through your commentary was that you might be able to extend the heuristic functionality to allow someone to implement a large response handler. For example, if the `Content-Length` is larger than some value (1GB maybe?) then you might trigger a custom buffer that avoids the extra reads / writes. 

I haven't looked at the code for a while, but I think heuristics only are for adjusting the response object before it is sent to the actual caching logic. You'd likely have to pass some extra info to trigger the buffering or potentially swap the response's body with your custom buffer. There are probably other colors to paint the bikeshed too!

I mention it as I suspect changing the core API might be challenging at this point whereas extending the heuristics might provide a more reasonable upgrade path over time.

hexagonrecursion:
> discrepancies between different operating systems

What kind of discrepancies?

> potential use cases

What kind of use cases?

> One idea I had when looking through your commentary was that you might be able to extend the heuristic functionality to allow someone to implement a large response handler. For example, if the `Content-Length` is larger than some value (1GB maybe?) then you might trigger a custom buffer that avoids the extra reads / writes.

Exactly how would this work? Where would the body be stored while it's being downloaded?

PR #242, opened 2021-02-19
Stop throwing if saving a request fails

hexagonrecursion:
Saving a cached request may fail for any number of reasons:
* Running out of memory
* Running out of disk space
* Other IO-related errors

Since we have already made the request and received the data we might
as well use it instead of dropping it on the floor just because we can't
save it to satisfy future requests.

This should make #145, #180 and #238 less severe.


hexagonrecursion:
Correction: this will not help with #145

hexagonrecursion:
@ionrock  Ping?

PR #247, opened 2021-02-23
Refactor: extract code to get responses from the cache

hexagonrecursion:
This PR extracts a helper method in `CacheController` to find a cached response and deserialize it.

This should make it easier to change the cache API. See https://github.com/ionrock/cachecontrol/pull/240

There is one place where I did not replace `cache.get()` with the new helper. It will be dealt with separately:
https://github.com/hexagonrecursion/cachecontrol/compare/refactor...hexagonrecursion:no-lbyl

### DIFF
--- a/cachecontrol/controller.py
+++ b/cachecontrol/controller.py
@@ -141,16 +141,8 @@ class CacheController(object):
             logger.debug('Request header has "max_age" as 0, cache bypassed')
             return False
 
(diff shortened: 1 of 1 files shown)


hexagonrecursion:
This PR and #240 will be closed on 11th or April 2021 due to lack of upstream feedback.

PR #252, opened 2021-08-12
Check for both max-age and s-maxage in the CacheControl header.

bob-schumaker:
I was banging my head against a wall trying to figure out why it wasn't caching and I found it :-)

The website I was calling used "s-maxage" in the CacheControl header, so now check for both.


frostming:
That isn't a standard header. We will not support this.

PR #256, opened 2021-10-14
Don't buffer all the data in memory when writing, reducing peak memory usage

### LINKED ISSUE #145: Large responses cause increased memory usage.

itamarst:
Fixes #145.

Before:

Git master as of 2021-10-14

After:

Code in PR

In both cases running the script from https://github.com/ionrock/cachecontrol/issues/145#issuecomment-293446396 using `fil-profile run example.py`.

### DIFF
--- a/cachecontrol/filewrapper.py
+++ b/cachecontrol/filewrapper.py
@@ -16,6 +16,13 @@ class CallbackFileWrapper(object):
 
     This class uses members with a double underscore (__) leading prefix so as
     not to accidentally shadow an attribute.
+
+    The data is stored in a temporary file until it is all available.  As long
+    as the temporary files directory is disk-based (sometimes it's a
+    memory-backed-``tmpfs`` on Linux), data will be unloaded to disk if memory
+    pressure is high.  For small files the disk usually won't be used at all,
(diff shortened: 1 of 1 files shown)


cachecontrol/filewrapper.py
```diff
@@ -62,8 +82,8 @@ def _close(self):
         # and allows the garbage collector to do it's thing normally.
         self.__callback = None
 
-        # Closing the BytesIO stream releases memory. Important when caching
-        # big files.
+        # Closing the tempoary file releases memory and frees disk space.
```
jjjamie, line 85: typo: s/tempoary/temporary/

ionrock: Thanks!

PR #264, opened 2021-10-31
Internal file handle revert

### LINKED ISSUE #263: Cache hits cause empty response content
After an upgrade of cachecontrol from 0.12.6 to 0.12.8 we experience empty response bodies on cache hits.

I've created a script that can reproduce the issue:
```
#!/usr/bin/env python3
import sys

import cachecontrol
import requests
from cachecontrol.cache import DictCache
from cachecontrol.heuristics import BaseHeuristic


class NoAgeHeuristic(BaseHeuristic):
    def update_headers(self, response):
        if "cache-control" in response.headers:
            del response.headers["cache-control"]


cache_adapter = cachecontrol.CacheControlAdapter(
    DictCache(), cache_etags=True, heuristic=NoAgeHeuristic()
)


session = requests.Session()
session.mount("https://", cache_adapter)

for i in range(2):
    response = session.get(
        "https://api.github.com/repos/sigmavirus24/github3.py/pulls/1033"
    )
    print(f'Content length: {len(response.content)}')
    if len(response.content) == 0:
        sys.exit(1)
```

When bisecting using the reproducer this issue seems to get introduced by https://github.com/ionrock/cachecontrol/commit/add2e5246130ca2738057aa6ff6d219cf763014d

ionrock:
We had a PR a while back that pre-emptively read the body and exhausted the internal file handle. This triggers a rare case where we need to read that file handle again apparently. 

The hope is that fixes #263.

### DIFF
--- a/cachecontrol/serialize.py
+++ b/cachecontrol/serialize.py
@@ -25,9 +25,15 @@ _default_body_read = object()
 
 
 class Serializer(object):
-
-    def dumps(self, request, response, body):
+    def dumps(self, request, response, body=None):
         response_headers = CaseInsensitiveDict(response.headers)
+
+        if body is None:
+            # When a body isn't passed in, we'll read the response. We
+            # also update the response with a new file handler to be
+            # sure it acts as though it was never read.
+            body = response.read(decode_content=False)
+            response._fp = io.BytesIO(body)
 
         # NOTE: This is all a bit weird, but it's really important that on
         #       Python 2.x these objects are unicode and not str, even when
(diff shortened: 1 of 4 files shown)


cachecontrol/serialize.py
```diff
@@ -25,10 +25,16 @@ def _b64_decode_str(s):
 
 
 class Serializer(object):
-
-    def dumps(self, request, response, body):
+    def dumps(self, request, response, body=None):
```
awilkins, line 28: Don't ned to revert this line, just explicitly pass `None` from `update_cached_response`?

Issue #265, opened 2021-11-03
Excessive memory usage, part 2


itamarst:
I'm testing new cachecontrol 0.12.8 with pip, and it only partially solves the excess memory usage: memory usage goes down from 1300MB to 900MB when doing `pip install tensorflow`.

The next bottleneck is the fact `msgpack` doesn't have a streaming interface, so packing always makes a copy of the data in memory. (I believe something like this was mentioned as a potential issue in the previous issues #145, but unfortunately the test script I was using didn't use that code path).

Some potential approaches:

1. Modify `msgpack` upstream to support streaming writes, perhaps coupled with streaming API on cachecontrol side.
2. Switch to a new data format where instead of having one giant bytestring, the msgpack is a series of bytestrings. This requires no changes to `msgpack`, and perhaps no public API changes assuming more `mmap` hackery.
3. Re-implement `msgpack` serialization just for this use case; essentially the code only packs a byte string, so just need to write the appropriate header beforehand. Depending on what APIs are public this may be the least intrusive option; hacky, but it'll work.
4. No doubt others (e.g. some other data format).

itamarst:
Related is https://github.com/pypa/pip/issues/9549, where someone tried to cache something bigger than msgpack message size. Solution 2 above would solve that too.

itamarst:
I will think about the best/easiest approach and submit a PR at some point soon, but if you have a preference/suggestion/other ideas, let me know.

ionrock:
@itamarst I don't have a strong opinion offhand. I suspect trying to get some changes into `msgpack` might be a long process. I'd wonder if there is a way to keep the metadata separate and just maintain a file manually to support the streaming? 

The other data format that I'm familiar with is Protobuf, but it doesn't have support for streaming after some quick research. There is Thrift to look into as well I suppose, but I'm not really sure if there are others to be considered.

itamarst:
Oh I hadn't read the part of the code where it actually constructs a dict with metadata. So it's not just the body in there...

itamarst:
There's... pickle. But that's a problematic format in many ways.

itamarst:
Updating as I learn:

CacheControl 0.12.6 memory usage was 3× size of cached value. 0.12.7 and later are 2×-ish: `msgpack` dump, and then an extra copy to add the version prefix. I say "ish" because there's an extra copy in a temporary file, and some of that may be dumped to disk if there is memory pressure.

[...]

PR #267, opened 2021-11-11
Work in progress, opened for initial API review: streaming APIs

itamarst:
This is a sketch of a fix for #265.

This is an initial API review; basically it allows streaming data to and from file-like objects, which means memory usage can be fixed instead of `O(2N)` as it is now.

It should be completely backwards compatible: at the moment all but a single test pass. The new APIs are not tested yet, this is more of a sketch, although they are used somewhat.

If you're happy with this approach I can do further work to turn this into tested, documented code.

### DIFF
--- a/cachecontrol/cache.py
+++ b/cachecontrol/cache.py
@@ -43,7 +43,7 @@ class CacheInterface(ABC):
     can support both.
     """
 
-    def get_as_file(self, key: str, serializer: Serializer) -> HTTPResponse:
+    def get_as_file(self, key: str, request, serializer: Serializer) -> HTTPResponse:
         """
         Return tuple of (encoded-metadata file-like, body file-like).
         """
--- a/cachecontrol/cache.py
+++ b/cachecontrol/cache.py
@@ -7,9 +7,13 @@ The cache object API for implementing caches. The default is a thread
 safe in-memory dictionary.
 """
 from threading import Lock
+from typing import Tuple, BinaryIO, Optional
+from pathlib import Path
+from abc import ABC
 
 
-class BaseCache(object):
+class LegacyCache(object):
+    """Legacy cache interface."""
 
     def get(self, key):
         raise NotImplementedError()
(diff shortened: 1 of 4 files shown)


itamarst:
@ionrock any thoughts?

itamarst:
@ionrock just checking in again. I don't want to put too much effort into making this actually releasable (tests, docs, etc.) if you're not happy with the proposed API.

ionrock:
@itamarst Thanks for digging into this. My only feedback is that changing the base cache seems pretty risky where it likely works well for a good majority of cases. I mentioned in my comment in #267 that it seems like we might want to focus on the serialization portion. For example, we're still calling serialize by default on the body in the `_cache_set` method, which seems like a good place to allow some logic to do the needful. I can imagine we could add a `body_handler` method to the BaseCache that when it exists is used in place of the serializer mechanics. That way you could take the body handler and save it in a file directly or whatever else (store in an object store for example). 

Let me know what you think or if you've run into issues that would not allow that to work. The handling of the body content has always been challenging, so it wouldn't surprise me if there are dragons in this tactic.

itamarst:
The problem is you can't make memory usage constant without changing the base cache API. You could drop the serialization, yes, and then go from `O(2N)` to `O(N)`: instead of two copies of the body, the original and the serialized, you'd only have one. So a 400MB download will go from needing 800MB of RAM to 400MB of RAM. Which is still an improvement, but it's still not ideal, 400MB is still a lot, and who knows what size downloads people are actually using.

The proposed design would be `O(1)`, probably a very small constant depending what size chunks are being received from the HTTP server. Presumably less than 1MB of RAM.

But if you're not comfortable with changes on this scale I can switch to the `O(N)` solution, since it is still an improvement.

itamarst:
Or. Hm. I guess given the `mmap` trick i might actually be possible to do `O(1)` memory usage if serialization goes away. It's a hack, but might work... let me think about it.

itamarst:
So looking through what you wrote above... adding a `body_handler` to `BaseCache` is indeed necessary for this to work with constant memory usage, which _is_ possible with `mmap` hack. Downsides:

1. That's still a change to the base cache class.
2. You end up writing two copies of the file to disk, which is annoying when they're hundreds of MB.
3. The interface is a little wonky in that you have two calls to the cache, which as someone doing implementation means more work, you need to deal with the intermediate step in a more complex way (e.g. there's no exception handler for errors as there would be for a single method). In practice... probably not a big deal.

It is a smaller change to the base cache class, though, and likely to the code base in general...

itamarst:
I think my design is better, but... the goal here is to optimize memory usage in one particular case, not to make this perfect, and current APIs work for most people I guess? So will try attempt that does something like this.

Issue #251, opened 2021-07-08
Document BaseCache expires value


thenewguy:
It would be helpful if the value for expires in `BaseCache.set` was documented.  I couldn't find it and the redis/file caches aren't explicit.

Is the expires value guaranteed to be in UTC?

Related #250

ionrock:
As the date is typically provided by the server, we parse the timezone in order to try to make sure we're making the appropriate comparisons. I can imagine it could be better to convert to UTC, but I can also imagine there could be dragons there too. 

In any case, if you look at the code and find a better answer that could use more docs and/or a change in behavior to use UTC, it seems like a legit change.

DawnofGenX:
I'd like to write this up (docs-only). Reading the code first so the docs state what actually happens rather than what the signature implies:

- The controller always passes an **int** (freshness lifetime in seconds: `max-age`, or `Expires - Date`), never a datetime, and only for entries still fresh at cache time.
- `expires` is purely a **backend hint**: `Serializer.dumps` never stores it — on lookup the controller recomputes freshness from the cached response headers. Custom backends must not rely on it being round-tripped.
- Backends differ today: `DictCache`/`FileCache` ignore it; `RedisCache` converts it to a TTL (naive datetime compared against naive local now, aware against UTC now).

Plan: a `BaseCache.set` docstring covering exactly that contract, plus a short "Cache Expiry" section in `docs/storage.rst` between the RedisCache and third-party-provider sections. No behavior changes. If the scope looks right I'll open a small PR shortly.

PR #269, opened 2021-11-29
Allow redis to set expires with an int arg

### LINKED ISSUE #268: Error when max-age is set
When using the redis cachecontrol, I get an error when the max-age header is set.

```
expires = 900

        if not expires:
            self.conn.set(key, value)
        else:
>           expires = expires - datetime.utcnow()
E           TypeError: unsupported operand type(s) for -: 'int' and 'datetime.datetime'

pyenv/lib/python3.6/site-packages/cachecontrol/caches/redis_cache.py:23: TypeError
```

I believe this is because we set `expires_time` to an `int` [here](https://github.com/ionrock/cachecontrol/blob/master/cachecontrol/controller.py#L349) instead of a `datetime`.

davegaeddert:
It seems like there are at least a couple scenarios in `controller.py` where the expires can be an int instead of a datetime.

https://github.com/ionrock/cachecontrol/blob/3bd694551d28445723c18f4ad8c44bce55ff9b5f/cachecontrol/controller.py#L325
https://github.com/ionrock/cachecontrol/blob/3bd694551d28445723c18f4ad8c44bce55ff9b5f/cachecontrol/controller.py#L349

With the way that works currently, I think some check needs to be done in `RedisCache` to see what type we're working with... the other solution that comes to mind is that `expires` should *only* be an int or a datetime? That would be a more involved change but I'd be willing to put some time into it @ionrock if you prefer it goes that direction. Implementing [mypy](https://github.com/python/mypy) or something might help sort out that kind of bug.

Fixes #268

### DIFF
--- a/tests/test_storage_redis.py
+++ b/tests/test_storage_redis.py
@@ -14,7 +14,7 @@ class TestRedisCache(object):
         self.conn = Mock()
         self.cache = RedisCache(self.conn)
 
-    def test_set_expiration(self):
+    def test_set_expiration_datetime(self):
         self.cache.set("foo", "bar", expires=datetime(2014, 2, 2))
         assert self.conn.setex.called
 


davegaeddert:
#251 (documentation around expires value) is also related to this.

ionrock:
@davegaeddert I think you're right about normalizing how we treat dates, but I also don't have a good idea of a direction to take. The ints seem better to work with overtime, but I could be convinced otherwise. If you're interested in trying to normalize things, I'd be grateful, but this change looks like a good start. Thanks!

davegaeddert:
Sounds good, thanks for merging it!

I'm not sure I know enough to push on normalizing... from some other issues it sounds like people also use custom cache backends and I'm not sure how to best navigate that.

PR #271, opened 2021-11-30
Reduce memory use, part 2

### LINKED ISSUE #265: Excessive memory usage, part 2

itamarst:
Fixes #265 

Tested memory usage the following program:

```python
import requests
from tempfile import mkdtemp
from cachecontrol import CacheControl
from cachecontrol.caches.file_cache import FileCache

cache = FileCache(mkdtemp())
sess = CacheControl(requests.Session(), cache=cache)
for chunk in sess.get("https://files.pythonhosted.org/packages/76/9a/a245290ae642f78caf80d1454045d19119dd6f266a81f02a1aa7aa89ba99/numpy-1.21.4-cp310-cp310-macosx_10_9_universal2.whl", stream=True).iter_content(16384):
    pass
```

Peak memory in master was 85MB (as tracked by Fil). In this PR, it was 7.7MB; I expect this to not change with file size. Given the downloaded file is 27MB, we've gone from `O(3N)` to `O(1)` memory usage.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -10,7 +10,7 @@
 0.12.11
 =======
 
-* ``FileCache`` now uses less memory.
+* Added new variant of ``FileCache``, ``SeparateBodyFileCache``, which uses less memory by storing the body in a separate file than metadata, and streaming data in and out directly to/from that file. Implemented by [Itamar Turner-Trauring](https://pythonspeed.com), work sponsored by [G-Research](https://www.gresearch.co.uk/technology-innovation-and-open-source/).
 
 0.12.7
 ======
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -7,7 +7,12 @@
  Release Notes
 ===============
 
-0.13.0
+0.12.11
+=======
+
+* ``FileCache`` now uses less memory.
+
+0.12.7
 ======
 
 * Dropped support for Python 2.7, 3.4, 3.5.
--- a/cachecontrol/cache.py
+++ b/cachecontrol/cache.py
@@ -52,6 +52,14 @@ class SeparateBodyBaseCache(BaseCache):
 
         cache.set(key, serialized_metadata)
         cache.set_body(key)
+
+    Similarly, the body should be loaded separately via ``get_body()``.
     """
     def set_body(self, key, body):
+        raise NotImplementedError()
+
+    def get_body(self, key):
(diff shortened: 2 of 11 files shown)


itamarst:
@ionrock when you have a moment, could you take a look?

docs/release_notes.rst
```diff
@@ -7,7 +7,12 @@
  Release Notes
 ===============
 
-0.13.0
+0.12.11
+=======
+
+* ``FileCache`` now uses less memory.
```
ionrock, line 13: I'd say leave yourself some more credit here while describing the approach :)

itamarst:
It occurs to me that this is not backwards-compatible as regards `FileCache` usage, e.g if people subclassed or wrapped it (which `pip` does).

So probably should have a new `MemoryEfficientFileCache` class or something, rather than modifying existing one.

itamarst:
OK, I have updated to separate class, added documentation, and improved release notes.

itamarst:
@ionrock what do you think?

itamarst:
@ionrock know it's a busy week, but if you have a few minutes, would love to get this merged or improved, depending.

itamarst:
If you happen to have any free time, would love a review/merge! (if not, no worries, I know how it goes).

itamarst:
Just checking in again, in case you have a few free minutes.

itamarst:
Thank you! Could you do a release next?

PR #272, opened 2021-12-01
Step #1 of release automation: Switch to setuptools_scm instead of bumpversion

itamarst:
This is a prerequisite for doing automatic releases on a GitHub tag.

In particular, instead of using `bumpversion`, the version is read from the latest Git tag. So to do a new release version, just add another Release in the GitHub UI for releases (a Release is a GitHub thing but it also automatically adds a corresponding tag).

At the moment version is reported as `0.12.7.dev60+g232132` because there is no tag for releases `0.12.7` to `0.12.10`, so if you choose to merge this (or probably even if you don't) you should add a tag for `0.12.10` at the very least.

There is also a new `RELEASE.md` document that documents the release process. A follow up PR will add automation to do the packaging and upload steps automatically.


RELEASE.md
```diff
@@ -0,0 +1,25 @@
...
+1. Make sure [release notes](docs/release_notes.rst) on main branch have latest changelog entries with version V.
+1. In GitHub UI, added release (or tag) with name `v<V>`, e.g. `v1.17.23`.
+
+## Upload
+
+At this point you build wheels and sdist:
+
+```
+$ python setup.py sdist
+$ pip wheel -w dist/ --no-deps .
```
frostming, line 15: After supporting automatic publishing, we should no longer need this step. But anyway, [build](https://pypi.org/project/build) is the preferred way to build distributions now.

docs/conf.py
```diff
@@ -45,16 +45,16 @@
 
 # General information about the project.
 project = u"CacheControl"
-copyright = u"2013, Eric Larson"
+copyright = u"2021, Eric Larson"
```
frostming, line 48: Why change this, the year means the person holds the copyright since 2013.

frostming (changes requested):
You may forget to remove the `.bumpversion.cfg`

PR #279, opened 2022-07-04
Add type annotations

dimbleby:
I probably should have checked first whether this is something you even want, but if the answer is "no thanks" then fair enough: it wasn't so much work that either of us need feel bad about that.

I've added type annotations everywhere, and mypy configuration to check them.  I haven't tried to update your github workflows, but all you should need to do is run like this:
```
$ mypy cachecontrol
Success: no issues found in 13 source files
```

### DIFF
--- a/cachecontrol/cache.py
+++ b/cachecontrol/cache.py
@@ -7,37 +7,39 @@ The cache object API for implementing caches. The default is a thread
 safe in-memory dictionary.
 """
 from threading import Lock
+from typing import TYPE_CHECKING, MutableMapping, Optional
+
+if TYPE_CHECKING:
+    from io import BufferedIOBase
 
 
 class BaseCache(object):
-
-    def get(self, key):
+    def get(self, key: str) -> Optional[bytes]:
         raise NotImplementedError()
 
-    def set(self, key, value, expires=None):
+    def set(self, key: str, value: bytes, expires: Optional[int] = None) -> None:
         raise NotImplementedError()
 
-    def delete(self, key):
+    def delete(self, key: str) -> None:
         raise NotImplementedError()
 
-    def close(self):
+    def close(self) -> None:
         pass
 
 
 class DictCache(BaseCache):
-
-    def __init__(self, init_dict=None):
+    def __init__(self, init_dict: Optional[MutableMapping[str, bytes]] = None) -> None:
         self.lock = Lock()
         self.data = init_dict or {}
 
-    def get(self, key):
+    def get(self, key: str) -> Optional[bytes]:
         return self.data.get(key, None)
 
-    def set(self, key, value, expires=None):
+    def set(self, key: str, value: bytes, expires: Optional[int] = None) -> None:
         with self.lock:
             self.data.update({key: value})
 
-    def delete(self, key):
+    def delete(self, key: str) -> None:
         with self.lock:
             if key in self.data:
                 self.data.pop(key)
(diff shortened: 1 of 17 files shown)


cachecontrol/caches/file_cache.py
```diff
@@ -62,12 +69,12 @@ class _FileCacheMixin:
...
-        forever=False,
-        filemode=0o0600,
-        dirmode=0o0700,
-        use_dir_lock=None,
-        lock_class=None,
+        directory: str,
+        forever: bool = False,
+        filemode: int = 0o0600,
+        dirmode: int = 0o0700,
+        use_dir_lock: Optional[bool] = None,
```
dimbleby, line 76: slightly curious API here, I've annotated it as it is.  

Suggest that `use_dir_lock: bool = False` might have done, though.

cachecontrol/compat.py
```diff
@@ -30,3 +30,5 @@
     text_type = unicode
 except NameError:
     text_type = str
+
+__all__ = ["urljoin", "pickle", "HTTPResponse", "is_fp_closed", "text_type"]
```
dimbleby, line 34: since setup.py requires recent python https://github.com/ionrock/cachecontrol/blob/6155c199438bfc3944a833f1dec6fb74134476f1/setup.py#L27

I suppose some or more of this file could be removed

cachecontrol/caches/file_cache.py
dimbleby: removed in 0.12.12 anyway

mr-c:
These are great! We had a very minimal set of type hints for cachecontrol; and I updated our type hints to match yours in https://github.com/common-workflow-language/schema_salad/commit/be8da2e23bd6e871b8f9cff93f955ddfcb52f95b ; and our type checking still pass.

One note would be to add a `mypy` test in `tox.ini`, so that future PRs are also checked.

cachecontrol/cache.py
```diff
@@ -7,37 +7,39 @@
...
+    def __init__(self, init_dict: Optional[MutableMapping[str, bytes]] = None) -> None:
         self.lock = Lock()
         self.data = init_dict or {}
 
-    def get(self, key):
+    def get(self, key: str) -> Optional[bytes]:
         return self.data.get(key, None)
 
-    def set(self, key, value, expires=None):
+    def set(self, key: str, value: bytes, expires: Optional[int] = None) -> None:
```
frostming, line 38: `expires` should also accept a datetime object, see the RedisCache implementation

cachecontrol/caches/file_cache.py
```diff
@@ -5,18 +5,24 @@
...
+
 try:
     FileNotFoundError
 except NameError:
     # py2.X
     FileNotFoundError = (IOError, OSError)
 
 
-def _secure_open_write(filename, fmode):
+def _secure_open_write(filename: str, fmode: int) -> "BufferedWriter":
```
frostming, line 25:
```suggestion
def _secure_open_write(filename: str, fmode: int) -> "IO[bytes]":
```
They work the same in typing, but `IO[bytes]` is simpler

cachecontrol/caches/file_cache.py
```diff
@@ -153,23 +160,23 @@ class SeparateBodyFileCache(_FileCacheMixin, SeparateBodyBaseCache):
     peak memory usage.
     """
 
-    def get_body(self, key):
+    def get_body(self, key: str) -> Optional["BufferedIOBase"]:
```
frostming, line 163:
```suggestion
    def get_body(self, key: str) -> Optional["IO[bytes]"]:
```

frostming (changes requested):

frostming (approved):

frostming (approved):

Issue #109, opened 2016-01-04
'lockfile' is deprecated.


dsully:
The 'lockfile' package required for file caching is deprecated.

https://pypi.python.org/pypi/fasteners is the recommended alternative.

sigmavirus24:
Can you provide a reference for it's deprecation?

dsully:
https://pypi.python.org/pypi/lockfile

"This package is deprecated".

ionrock:
@dsully Thanks for bringing this up. Did you run into an issue with lockfile that is fixed with fasteners?

dsully:
@ionrock: No, I ran into a dependency issue regarding 'pbr', which is declared as setup_requires for lockfile, but is actually used at package install time.

I notice that there is no equivalent to lockfile's MkDirLockFile class as well.

dstufft:
I attempted to implement this, but I ran into an issue where it doesn't appear to actually be locking so I punted on it for the time being. However handling this in a compatible way might be tricky. There are a few scenarios that someone could be in currently:

1. They don't care about the type of lock used at all and just used the defaults.
2. They wanted to use directory locking and passed in ``use_dir_locking``.
3. They specifically wanted to use a very particular lock class and passed that in specifically.

For the people in group 1, the main backwards compatibility concerns is whether or not cachecontrol wants to support compatibility with people running older and newer versions of the code at the same time. If it does not, then great, you can just switch to fasteners and not think about it. However if you do, then you'll need to be concerned about how to ensure that the existing lock file actually continues to cause a lock to happen, even with the new code.

For the people in group 2, you either have to remove this parameter or continue to support pylockfile when this parameter is chosen. This one is easy because you either continue to use the old code, or you just stop supporting it. However if fasteners ever gets something like ``MkDirLockFile`` then the same issue occurs as with group 1.

For the people in group 3, you don't have a lot of specific compatibility concerns as long as you ensure the lock API that you're using is compatible amongst both pylockfile and fasteners or you drop support for using specific pylockfile classes.

[...]

PR #284, opened 2022-09-10
switch from lockfile to filelock

dimbleby:
I see that this project is rather inactive, but I guess I'm an optimist...

lockfile has long been deprecated, per https://github.com/ionrock/cachecontrol/issues/109 (2016!). https://github.com/ionrock/cachecontrol/pull/114 was an earlier try at getting cachecontrol onto something that was both active and better, it went with fasteners. These days filelock is getting about ten times as many downloads as fasteners - so I've used that.

Among the ways that this fixes actual bugs - ie is not just change for its own sake - is that the mechanisms used by filelock will automatically release the lock if the process exits early eg by Ctrl-C or bug or whatever.

### DIFF
--- a/cachecontrol/caches/file_cache.py
+++ b/cachecontrol/caches/file_cache.py
@@ -66,32 +66,22 @@ class _FileCacheMixin:
         forever=False,
         filemode=0o0600,
         dirmode=0o0700,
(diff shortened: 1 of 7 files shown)


ionrock:
Thanks!

Issue #282, opened 2022-09-02
Tests find their way into wheels


KOLANICH:

frostming:
Closed by #287

PR #287, opened 2022-09-13
Packaging the tests, or not

### LINKED ISSUE #281: No tests in pypi sdist tarball
Hi! I'm packaging this project for Arch Linux.

We usually try to rely on pypi.org for sdist tarballs to build from. In a packaging context we also run tests to ensure, that the given project is compatible with our own Python ecosystem. Unfortunately the tests are not included in the sdist tarball on pypi.org, so it would be awesome if you could add them there. It allows anyone to also rely on tests when they build a wheel from source! :)

dimbleby:
Fixes #281, #282

- first commit explicitly excludes tests from the binary distribution, it is indeed unconventional to include them
- second commit explicitly re-adds them to the source distribution
  - I don't think there's universal consensus about whether this is desirable or not eg [here](https://discuss.python.org/t/should-sdists-include-docs-and-tests/14578) is a recent discussion
  - but it seems harmless and seeing as how #281 explicitly asks for them, well why not?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+graft tests/
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup_params = dict(
     author_email="eric@ionrock.org",
     url="https://github.com/ionrock/cachecontrol",
     keywords="requests http caching web",
-    packages=setuptools.find_packages(),
+    packages=setuptools.find_packages(exclude=["tests", "tests.*"]),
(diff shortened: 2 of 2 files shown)


itamarst:
This would enable re-releasing the yanked .12 release, seems like a good idea.

itamarst:
Any chance this can be merged and a new release made? Thanks!

KOLANICH:
Thanks for fixing that.

PR #289, opened 2022-11-12
Use Python's unittest mock

jelly:
Since 3.3 Python unittest module supports mock and the seperate mock module is deprecated.


frostming:
Thanks

PR #290, opened 2023-04-30
Remove use of utcnow

pganssle:
`utcnow` [will be deprecated in Python 3.12](https://github.com/python/cpython/issues/103857). In anticipation of this, it would be good to [switch over your use of `datetime.utcnow` to use actual timezone-aware datetimes](https://blog.ganssle.io/articles/2019/11/utcnow.html).

There are two commits in this PR. The first one replaces all uses of `utcnow` with drop-in equivalents, but doesn't solve the underlying problem. The second one converts all the use of `datetime` that I found to using aware datetimes instead of naïve ones. I think accepting that one would be preferable.

### DIFF
--- a/cachecontrol/heuristics.py
+++ b/cachecontrol/heuristics.py
@@ -4,7 +4,7 @@
 
 import calendar
 import time
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from email.utils import formatdate, parsedate, parsedate_tz
 from typing import TYPE_CHECKING, Any, Dict, Mapping, Optional
 
@@ -15,7 +15,7 @@ TIME_FMT = "%a, %d %b %Y %H:%M:%S GMT"
 
 
 def expire_after(delta: timedelta, date: Optional[datetime] = None) -> datetime:
-    date = date or datetime.utcnow()
+    date = date or datetime.now(timezone.utc).replace(tzinfo=None)
     return date + delta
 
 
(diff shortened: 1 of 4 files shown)


pganssle:
One thing to note: this is the only library that `pip` vendors that uses `utcnow`. Ideally we would get this taken care of ASAP so that it can be vendored in `pip` and have `utcnow` removed from `pip` entirely as soon as possible, since ideally `pip` would be prepared for the deprecation well in advance of the 3.12 release.

frostming:
@pganssle Can you please rebase the master?

tests/test_expires_heuristics.py
```diff
@@ -185,7 +187,8 @@ def test_last_modified_is_used_when_cache_control_public(self):
         )
         modified = self.heuristic.update_headers(resp)
         assert ["expires"] == list(modified.keys())
-        assert datetime(*parsedate(modified["expires"])[:6]) > datetime.now()
+        expected = datetime(*parsedate(modified["expires"][:6]), tzinfo=timezone.utc)
```
frostming, line 190:
```suggestion
        expected = datetime(*parsedate(modified["expires"])[:6], tzinfo=timezone.utc)
```

cachecontrol/heuristics.py
```diff
@@ -67,7 +67,7 @@ def update_headers(self, response: "HTTPResponse") -> Dict[str, str]:
 
         if "expires" not in response.headers:
             date = parsedate(response.headers["date"])
-            expires = expire_after(timedelta(days=1), date=datetime(*date[:6]))
+            expires = expire_after(timedelta(days=1), date=datetime(*date[:6], tzinfo=timezone.utc))
```
frostming, line 70:
```suggestion
            expires = expire_after(timedelta(days=1), date=datetime(*date[:6], tzinfo=timezone.utc))  # type: ignore[misc]
```

frostming (approved):

PR #294, opened 2023-05-05
fix: compatibility with urllib3 2.0

### LINKED ISSUE #292: cachecontrol is incompatible with Requests 2.30.0
See https://github.com/psf/requests/issues/6437

### LINKED ISSUE #293: serialize.py function fails with HTTPResonse has no attribute "strict"
When running `pdm install` with no dependencies on python 3.10 and 3.11 I have just started experiencing at error stemming from [this line](https://github.com/ionrock/cachecontrol/blob/c05ef9eff1c9ac176481fb99e7a7188aa5b4e17b/cachecontrol/serialize.py#L54). I don't have a lot of information beyond this since this is for some reason called by `pdm sync/install`. I do know, that replacing the `response.strict` part with `True` got `pdm sync` to stop crashing.

frostming:
Signed-off-by: Frost Ming <me@frostming.com>

Fix #292 
Fix #293

### DIFF
--- a/cachecontrol/serialize.py
+++ b/cachecontrol/serialize.py
@@ -51,7 +51,6 @@ class Serializer(object):
                 u"status": response.status,
                 u"version": response.version,
                 u"reason": text_type(response.reason),
-                u"strict": response.strict,
                 u"decode_content": response.decode_content,
             }
         }
--- a/tests/test_cache_control.py
+++ b/tests/test_cache_control.py
@@ -5,18 +5,19 @@
 """
 Unit tests that verify our caching methods work correctly.
 """
-import pytest
-from mock import ANY, Mock
 import time
 from tempfile import mkdtemp
+
+import pytest
+from mock import ANY, Mock
 
 from cachecontrol import CacheController
 from cachecontrol.cache import DictCache
 from cachecontrol.caches import SeparateBodyFileCache
-from .utils import NullSerializer, DummyResponse, DummyRequest
+
+from .utils import DummyRequest, DummyResponse, NullSerializer
 
 TIME_FMT = "%a, %d %b %Y %H:%M:%S GMT"
-
 
 
 class TestCacheControllerResponse(object):
@@ -126,13 +127,16 @@ class TestCacheControllerResponse(object):
         cache = DictCache({})
         cc = CacheController(cache)
         req = DummyRequest(url="http://localhost/", headers={"if-match": "xyz"})
-        resp = DummyResponse(status=304, headers={
-            "ETag": "xyz",
-            "x-value": "b",
-            "Date": time.strftime(TIME_FMT, time.gmtime()),
-            "Cache-Control": "max-age=60",
-            "Content-Length": "200"
-        })
+        resp = DummyResponse(
+            status=304,
+            headers={
+                "ETag": "xyz",
+                "x-value": "b",
+                "Date": time.strftime(TIME_FMT, time.gmtime()),
+                "Cache-Control": "max-age=60",
+                "Content-Length": "200",
+            },
+        )
         # First, ensure the response from update_cached_response() matches the
         # cached one:
         result = cc.update_cached_response(req, resp)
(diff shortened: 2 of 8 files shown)


adiroiban:
Thanks @frostming for the change.

I see that besides the `strict` keyword removal this PR also does import reordering and some reformating and modernization changes.

----------

Also for older urllib versions, this will just ignore the strict argument.

I think that we should wait for @ionrock to review this.

I was expecting to remove `strict` only for newer urllib versions, so that cachecontrol will continue to work with both old and new urllib versions.

ionrock:
I agree with @adiroiban that we should avoid formatting changes. The caveat is unless black reformats something, in which case two commits might be helpful to communicate that is what is happening. I'd also skip the `setup` to `setup_method` as it isn't necessary.

I also think it is worthwhile to handle the earlier urllib versions. When I started looking at it, I had considered a function in `compat.py` that checked the version and would remove the `strict` field there based on the version. 

Thanks for diving into this!

miigotu:
Why not just set the minimum version of urllib supported by the next version, so that earlier urllib versions will just use the older cachecontrol? Trying to be backwards compatible with an obsolete package version seems unnecessary when a working version of cachecontrol already exists for those older versions. Sort of what versioning and pinning is for imo.

frostming:
> Why not just set the minimum version of urllib supported by the next version

I agree with this strategy. Requesting a library to support all possible Python versions and urllib3 versions is a pain for maintainers. Also we seem to have to drop Python 3.6 in the CI matrix.

adiroiban:
Good feedback.

Only supporting the latest urllib3 is a valid point, but then, this PR should update `setup.py` and declare this explicit dependency.

```
install_requires=["requests", "msgpack>=0.5.2"],
```

Thanks

jacobg:
Hi, any status update on this one?

frostming:
Hi, I have to move on to create a fork at https://github.com/frostming/cacheyou, with this issue addressed.

The long term plan is if this project can continue I will merge the changes back.

dairiki:
> I also think it is worthwhile to handle the earlier urllib versions. When I started looking at it, I had considered a function in `compat.py` that checked the version and would remove the `strict` field there based on the version.

Note that (when running under Python 3) `urllib3==1.*` simply [discards and ignores](https://github.com/urllib3/urllib3/blob/3c0148048a523325819377b23fc67f8d46afc3aa/src/urllib3/connection.py#L116-L117) any `strict` parameter passed to the HTTPConnection constructor.  It has been this way since `urllib3==1.8` (circa 2014, see urllib3/urllib3#343).  (It is also [noted in the urllib3 docs](https://github.com/urllib3/urllib3/blob/3c0148048a523325819377b23fc67f8d46afc3aa/src/urllib3/connectionpool.py#L124-L130) that that `strict` has no effect outside of Python 2.)

Since `cachecontrol` declares `Requires-Python: >=3.6`, it seems that it is pretty safe to simply remove the `strict` field.

miigotu:
> Hi, I have to move on to create a fork at https://github.com/frostming/cacheyou, with this issue addressed.
> 
> The long term plan is if this project can continue I will merge the changes back.

Thanks for that, I was close to doing the same thing but I'm glad someone stepped up for it. Sickchill is using your fork for the foreseeable future since cachecontrol hasn't been updated in years. I hope you can keep it relevant and set up a nice GitHub actions workflow to keep it always maintained and supporting the latest versions of everything.

miigotu:
> > I also think it is worthwhile to handle the earlier urllib versions. When I started looking at it, I had considered a function in `compat.py` that checked the version and would remove the `strict` field there based on the version.
> 
> Note that (when running under Python 3) `urllib3==1.*` simply [discards and ignores](https://github.com/urllib3/urllib3/blob/3c0148048a523325819377b23fc67f8d46afc3aa/src/urllib3/connection.py#L116-L117) any `strict` parameter passed to the HTTPConnection constructor.  It has been this way since `urllib3==1.8` (circa 2014, see urllib3/urllib3#343).  (It is also [noted in the urllib3 docs](https://github.com/urllib3/urllib3/blob/3c0148048a523325819377b23fc67f8d46afc3aa/src/urllib3/connectionpool.py#L124-L130) that that `strict` has no effect outside of Python 2.)
> 
> Since `cachecontrol` declares `Requires-Python: >=3.6`, it seems that it is pretty safe to simply remove the `strict` field.
> 
> [ETA: Also note that the `strict` parameter hasn't been supported by `http.client.HTTPResponse` [since before Python 3.2](https://github.com/python/cpython/blob/29360893df0931703cb2fe0c0fbc0fc963e64ead/Lib/http/client.py#L283-L286).]
> 
> [ETA: IOW, the `strict` parameter, even where accepted, doesn't actually *do* anything under any extant version of Python 3. AFAICT, it can be safely ignored/dropped unless the intent is to support Python 2.]

Just supporting the latest on python 3.8+ is ideal, backwards compatibility was a huge importance a few years ago before most projects finally got out of the python 2.7 trap. With GitHub actions and all the great tools now almost everyone has been able to keep their dependency trees updated thanks to dependabot. Older versions will just get the older cachecontrol with the pip dependency resolver so nobody misses anything to drop backwards compatible.

PR #308, opened 2023-06-01
workflows: add a release workflow

woodruffw:
This adds a release workflow that runs on a `release: published` event, with both PyPI uploading (via trusted publishing) and code signing (via Sigstore). I kept the latter because I copied this from some other projects that I use the same workflow on, but I'm happy to remove it in the interest of keeping things minimal to begin with.

Also, correcting myself from earlier: it looks like I don't have permissions to register the trusted publisher on the PyPI side:

Given that, I'll mark this as blocked until @ionrock can register the publisher (or grant me or @frostming the ability to do it ourselves).


.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - uses: actions/setup-python@v4
+      with:
+        python-version: ">= 3.6"
+
+    - name: deps
+      run: python -m pip install -U setuptools build wheel
```
frostming, line 29:
```suggestion
      run: python -m pip install -U build
```

woodruffw: Good catch, thanks.

frostming:
Maybe add a `pyproject.toml` to explicitly specify the build backend?

https://packaging.python.org/en/latest/tutorials/packaging-projects/#creating-pyproject-toml

woodruffw:
> Maybe add a `pyproject.toml` to explicitly specify the build backend?

Sounds good. Want me to move the whole `setup.py`/`setup.cfg` over while I'm at it?

frostming:
> Sounds good. Want me to move the whole `setup.py`/`setup.cfg` over while I'm at it?

Sure, if the config can be read from `pyproject.toml`

pyproject.toml
```diff
@@ -0,0 +1,74 @@
+[build-system]
+requires = ["flit_core >=3.2,<4"]
+build-backend = "flit_core.buildapi"
+
+[tool.flit.module]
+name = "cachecontrol"
```
woodruffw, line 6: I went with `flit` because it's what I've used it in the past, but I'm happy to use another build backend (like setuptools' own backend) if you'd prefer.

pyproject.toml
```diff
@@ -0,0 +1,74 @@
...
+    "Topic :: Internet :: WWW/HTTP",
+]
+keywords = ["requests", "http", "caching", "web"]
+dependencies = ["requests >= 2.16.0", "msgpack >= 0.5.2"]
+requires-python = ">=3.6"
+
+[project.urls]
+Homepage = "https://pypi.org/project/CacheControl/"
+Issues = "https://github.com/psf/cachecontrol/issues"
+Source = "https://github.com/psf/cachecontrol"
```
woodruffw, line 39: Also went ahead and updated + added to the project's URLs.

pyproject.toml
frostming: I remember `flit build` and `python -m build` produce different distribution files. Can you verify all required files are included in the sdist and wheel by running `python -m build`?

woodruffw: The wheel looks correct to me, comparing against [this](https://inspector.pypi.io/project/cachecontrol/0.13.0/packages/65/03/9ed07561cc4b428204fc59fa090ae6652b6fe0a6506c2b05adfcb09085b9/CacheControl-0.13.0-py3-none-any.whl/).

The sdist also looks "correct," (from [this](https://inspector.pypi.io/project/cachecontrol/0.13.0/packages/9d/47/9bdf134012537324bbdebe84c144cc8fab08f148fc6c9ce89b8eefff1cb0/CacheControl-0.13.0.tar.gz/)) although now it's missing the `tests` module. I can try and re-add that if you'd like, although I'm not sure if it's considered idiomatic anymore to include the testsuite in the sdist.

woodruffw: I also confirmed that both the wheel and sdist install correctly (via `python -m pip install dist/....`) and pass the local dev copy's tests.

frostming: Yes, include the `tests` for sdist please.

woodruffw: Sure thing, done with f3e4222.

frostming (approved):

ionrock:
@woodruffw Done!

ionrock:
@woodruffw I'm assuming you're going to add the `release.yml` as it isn't in the workflows folder.

woodruffw:
Thanks @ionrock!



> @woodruffw I'm assuming you're going to add the `release.yml` as it isn't in the workflows folder.

Yep, it should be included in this PR's changeset: https://github.com/psf/cachecontrol/pull/308/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34

woodruffw:
This should be good to go whenever. @frostming, I assume we'll want to merge your 3.7 changes first.

woodruffw:
@frostming I'll do a pre-release to test the CI here after this is merged, unless you have any objections :slightly_smiling_face:

frostming:
@woodruffw Go ahead, let's also update the Makefile and include a dev guide for releasing.

frostming:
@ionrock Any chance to yank 0.13.0 to prevent users on Python 3.6 from installing this version(yeah we didn't know the context of yanking 0.12.12).

@woodruffw If the version is yanked we should be able to releae this with 0.13.1

ionrock:
@frostming Yanked!

woodruffw:
Okay, pushed up some initial contributing docs (in the `CONTRIBUTING.md` style that GitHub encourages). 

I've removed the old `make bump` target since `bumpversion` is unmaintained and doesn't have great support for pre-releases anyways. I'll look into other bumping tools, but for the mean time I've documented a "manual" bumping flow that's almost as simple (since I've consolidated all versions to a single file).

frostming:
@woodruffw what about https://pypi.org/project/bumpver which should be a drop-in replacement for `bumpversion`

woodruffw:
Works for me -- I'll look at adding it tomorrow.

PR #310, opened 2023-06-01
serialize: remove v1 and v2 formats

### LINKED ISSUE #309: RFC: Remove `serialize._loads_v1` and `_loads_v2`
Opening this as an RFC because it's technically a behavioral change.

At the moment, CacheControl has 5 serialization "implementations":

* `_loads_v0`: legacy, currently a no-op
* `_loads_v1`: legacy, loads from a pickle file
* `_loads_v2`: legacy, loads from a zlib-compressed base64 stream
* `_loads_v3`: legacy, currently a no-op
* `_loads_v4`: the current msgpack-based serialization

These are all defined under `cachecontrol.serialize`, and are dispatched to based on the version in the cache entry's "header". 

I propose removing `_loads_v1` and `_loads_v2` in a similar manner to v0/v3: we can just make them no-ops, meaning that any entries that happen to be found in those formats will be treated as cache-misses instead. 

Rationale:

* For just `_load_v1`: loading from pickle files is well known to be [dangerous](https://docs.python.org/3/library/pickle.html).
* For both: both formats are very old, and were deprecated years ago (around 2017, with `0.12.0`). As a result, any lingering cache entries that happen to be on still-running systems are very likely expired anyways.

Benefits:

* For `_load_v1`, removing an old and potentially dangerous deserialization surface
* For both: generally reducing the complexity of the de/serialization code, and allowing us to simplify the `serialize.loads` lookup logic

Thoughts? If this sounds reasonable, I'm happy to do it.

woodruffw:
Closes #309.

### DIFF
--- a/cachecontrol/serialize.py
+++ b/cachecontrol/serialize.py
@@ -2,11 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import base64
 import io
-import json
-import pickle
-import zlib
 from typing import IO, TYPE_CHECKING, Any, Mapping, Optional
 
 import msgpack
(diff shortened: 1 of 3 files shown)


frostming:
I think it may be a good idea to refactor a bit to extract the serialization/deserialization logic to its own class, instead of mixing them together. We can do it in the following PRs.

woodruffw:
> I think it may be a good idea to refactor a bit to extract the serialization/deserialization logic to its own class, instead of mixing them together. We can do it in the following PRs.

Yeah, sounds good. I need to do a little more cleanup in this PR but it should be ready in a moment.

frostming (approved):

PR #312, opened 2023-06-07
feat: drop support of Python older than 3.7

frostming:
Signed-off-by: Frost Ming <me@frostming.com>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,11 +18,6 @@ jobs:
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
         os: ["macos-latest", "windows-latest", "ubuntu-latest"]
-        include:
-          - python-version: "3.6"
-            os: "ubuntu-20.04"
-          - python-version: "3.6"
-            os: "windows-latest"
 
     steps:
       - uses: "actions/checkout@v3"
--- a/cachecontrol/_cmd.py
+++ b/cachecontrol/_cmd.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: 2015 Eric Larson
 #
 # SPDX-License-Identifier: Apache-2.0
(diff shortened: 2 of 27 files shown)


woodruffw (approved):
LGTM!

woodruffw:
> we have to make such change in a 0.14.0 release

That's fine by me -- IMO we _technically_ can make it with a patch as well since we're pre-1.0, but I don't have any preference and doing it with `0.14.0` is probably clearer.

dimbleby:
python 3.7 goes eol at the end of the month - it would not be unreasonable to drop support for that too in the next release

woodruffw:
> python 3.7 goes eol at the end of the month - it would not be unreasonable to drop support for that too in the next release

One of CacheControl's major users is `pip`, so we probably need to be _slightly_ more conservative than Python's own EOL schedule.

The current `pip` support policy is to support any Python version with more than 5% of PyPI's total downloads: https://pip.pypa.io/en/stable/development/release-process/#python-support-policy

dimbleby:
Of course whether you want to continue to support python 3.7 is up to maintainers here: I just wanted to check that you weren't about to drop python 3.6 and then a week later go "oops, we really ought to have dropped python 3.7 at the same time".

FWIW it seems to me that pip's policies are pip's problem!  but if you all don't find it burdensome to go along with them then all power to you.

woodruffw:
> FWIW it seems to me that pip's policies are pip's problem! but if you all don't find it burdensome to go along with them then all power to you.

Agreed 🙂 -- I see it as a balance between courtesy for a downstream and our own development experience. The jump from 3.6 to 3.7 is pretty big in terms of `__future__` support and type hint functionality whereas 3.7 to 3.8 is a bit more incremental, so I'm (personally) okay with not making more hops.

frostming:
Agree, we won't drop Python 3.7 shortly.

PR #323, opened 2023-09-15
Add support for Python 3.12 and drop EOL 3.6 and 3.7

hugovk:
The [second Python 3.12 release candidate](https://discuss.python.org/t/python-3-12-0rc2-final-release-candidate-released/33105/5?u=hugovk) is out! :rocket:

> ## Call to action
> 
> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.12 compatibilities during this phase, and where necessary publish Python 3.12 wheels on PyPI to be ready for the final release of 3.12.0.

Python 3.12.0 final is due out in two weeks: [2023-10-02](https://peps.python.org/pep-0693/).

See also https://dev.to/hugovk/help-test-python-312-beta-1508/

---

Python 3.7 is EOL and no longer receiving security updates (or any updates) from the core Python team.

| version |  released  |    eol     |
|:--------|:----------:|:----------:|
| 3.11    | 2022-10-24 | 2027-10-24 |
| 3.10    | 2021-10-04 | 2026-10-04 |
| 3.9     | 2020-10-05 | 2025-10-05 |
| 3.8     | 2019-10-14 | 2024-10-14 |
| 3.7     | 2018-06-26 | 2023-06-27 |

https://devguide.python.org/versions/

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
       contents: write
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - uses: actions/setup-python@v4
       with:
@@ -35,7 +35,7 @@ jobs:
       uses: pypa/gh-action-pypi-publish@release/v1
 
     - name: sign
-      uses: sigstore/gh-action-sigstore-python@v1.2.3
+      uses: sigstore/gh-action-sigstore-python@v2.0.1
       with:
         inputs: ./dist/*.tar.gz ./dist/*.whl
         release-signing-artifacts: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,7 +23,7 @@ jobs:
 
     - uses: actions/setup-python@v4
       with:
-        python-version: ">= 3.6"
+        python-version: "3.x"
 
     - name: deps
       run: python -m pip install -U build
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,14 +16,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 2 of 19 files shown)


woodruffw:
Thanks @hugovk! This LGTM; I'll wait for @frostming to additionally OK the Python version deprecations here.

.github/workflows/release.yml
```diff
@@ -32,10 +32,10 @@ jobs:
       run: python -m build
 
     - name: publish
-      uses: pypa/gh-action-pypi-publish@release/v1
+      uses: pypa/gh-action-pypi-publish@v1.8.10/v1
```
woodruffw, line 35: I _think_ we want to keep this as `release/v1`, rather than using a specific tag.

hugovk: Ah yes, this was a mistake. I have a script to update tags, but it doesn't handle this one very well. I'll revert it.

dvershinin:
@hugovk just chiming it to say that not everything that is deprecated should be quickly discarded as unused. There are LTS operating system versions which are based off Python 3.6 which *are* maintained, e.g. RHEL 8 is maintained at least till 2029.

woodruffw (approved):

woodruffw:
> @hugovk just chiming it to say that not everything that is deprecated should be quickly discarded as unused. There are LTS operating system versions which are based off Python 3.6 which are maintained, e.g. RHEL 8 is maintained at least till 2029.

I'm not sure this is a great reason for us, as an upstream, to continue supporting EOL Python versions: maintaining (and selling) an LTS Linux distribution comes with the expectation of carrying patches and support potentially past the point of upstream support. 

Given that and the fact that 3.6 and 3.7 are sitting at less that 5% total downloads _combined_ (https://pypistats.org/packages/cachecontrol), my first inclination is to say that we should drop these. That being said, Requests itself also has a minimum Python version of 3.7, so maybe we should piggyback on their policy and drop only 3.6 for now. Thoughts @hugovk?

hugovk:
Sure, updated!

woodruffw (approved):

woodruffw:
Thanks @hugovk!

PR #326, opened 2024-01-29
Fix document code case errors

somata-cn:
In the SeparateBodyFileCache chapter, there is an error in the provided document sample, and the code cannot be executed correctly


frostming (approved):

woodruffw (approved):
Thank you @somata-cn!

PR #328, opened 2024-02-01
Don't use the cache if the request has a Range header

### LINKED ISSUE #327: download of partial content is defeated by cachecontrol
suppose some large file is in the cache already and we make a request for partial content with a `Range` header.

understanding that caching of such things is likely to go wrong we also include the header `Cache-Control: no-cache`.

this header succeeds in preventing cachecontrol from responding immediately from the cache - but does not stop it from going through https://github.com/psf/cachecontrol/blob/7e3edbf61fa8d6cd2db1a0eed3a2e1c2fbc67d18/cachecontrol/adapter.py#L74 in which it unconditionally looks in the cache https://github.com/psf/cachecontrol/blob/7e3edbf61fa8d6cd2db1a0eed3a2e1c2fbc67d18/cachecontrol/controller.py#L273.  The result is that cachecontrol adds an `If-Modified-Since` header to the request, with timestamp per what is in the cache.

but now the server sees the header on that request, knows that the file has not been modified recently, and returns 304.  And now cachecontrol returns the entire file from its cache anyway, and not the partial content that the server would have given.

related-ish: #246 asks for cachecontrol to cache partial content.  But this one reports that cachecontrol makes partial content not work

dimbleby:
fixes #327

### DIFF
--- a/cachecontrol/controller.py
+++ b/cachecontrol/controller.py
@@ -142,6 +142,11 @@ class CacheController:
         """
         Load a cached response, or return None if it's not available.
         """
+        # We do not support caching of partial content: so if the request contains a
+        # Range header then we don't want to load anything from the cache.
+        if "Range" in request.headers:
+            return None
+
         cache_url = request.url
         assert cache_url is not None
         cache_data = self.cache.get(cache_url)
--- a/tests/test_etag.py
+++ b/tests/test_etag.py
@@ -79,6 +79,31 @@ class TestETag:
         # Make sure we updated our cache with the new etag'd response.
         assert self.cache.get(self.etag_url) == resp.raw
 
+    def test_etags_get_no_cache(self, sess, server):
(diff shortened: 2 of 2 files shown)


woodruffw (approved):
Thanks @dimbleby, LGTM. I'll let @frostming review as well before merging.

(Also, thanks for adding a test!)

frostming:
Looks good for now. I think we would have to encode headers as well as the URL as the cache key.

woodruffw:
> I think we would have to encode headers as well as the URL as the cache key.

Makes sense!

@dimbleby: Do you need a point release for this fix? If so, I can prep one tonight.

frostming:
> @dimbleby: Do you need a point release for this fix? If so, I can prep one tonight.

I think so, we can cut a `0.14.0` release

dimbleby:
thanks folks!

PR #333, opened 2024-04-27
GitHub Action to spellcheck and lint Python code

cclauss:
This Action uses minimal steps to run in ~5 seconds to rapidly:
* use `codespell` to look for typos in the codebase, and
* use `ruff` to lint Python code and provide intuitive GitHub Annotations to contributors.

Tools:
* https://github.com/codespell-project/codespell#readme
* https://docs.astral.sh/ruff

% `codespell --write-changes`
```
./docs/index.rst:69: sucessfully ==> successfully
./docs/tips.rst:46: tranferring ==> transferring
./docs/etags.rst:21: presense ==> presence
./cachecontrol/filewrapper.py:41: vaguaries ==> vagaries
./cachecontrol/filewrapper.py:44: thigns ==> things
./tests/test_cache_control.py:178: wih ==> with
```
% `ruff check --target-version=py38`
```
Error: cachecontrol/caches/file_cache.py:9:39: F401 `typing.Union` imported but unused
Error: tests/conftest.py:163:5: E722 Do not use bare `except`
Error: tests/issue_263.py:16:1: E402 Module level import not at top of file
Error: tests/issue_263.py:16:20: F401 `pprint.pprint` imported but unused
Error: tests/test_serialization.py:5:8: F401 `pickle` imported but unused
Error: Process completed with exit code 1.
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,7 +21,6 @@ jobs:
         exclude:
           - python-version: "3.7"
             os: "macos-latest"
-            
 
     steps:
       - uses: "actions/checkout@v4"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,9 +18,6 @@ jobs:
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
         os: ["macos-latest", "windows-latest", "ubuntu-latest"]
-        exclude:
-          - python-version: "3.7"
-            os: "macos-latest"
 
     steps:
       - uses: "actions/checkout@v4"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,6 +18,10 @@ jobs:
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
         os: ["macos-latest", "windows-latest", "ubuntu-latest"]
+        exclude:
+          - python-version: "3.7"
+            os: "macos-latest"
+            
 
     steps:
       - uses: "actions/checkout@v4"
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ bootstrap: $(VENV)/bin/pip
 
 format:
 	$(VENV)/bin/codespell
-	$(VENV)/bin/ruff check
+	$(VENV)/bin/ruff check --fix
 	$(VENV)/bin/ruff format
 
 doc: $(VENV)/bin/sphinx-build
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,9 @@ bootstrap: $(VENV)/bin/pip
 	$(VENV)/bin/pip install -e .[dev]
 
 format:
-	$(VENV)/bin/black .
+	$(VENV)/bin/codespell
+	$(VENV)/bin/ruff check
+	$(VENV)/bin/ruff format
 
 doc: $(VENV)/bin/sphinx-build
 	. $(ACTIVATE);
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,5 +79,3 @@ ignore_missing_imports = true
 [tool.pytest.ini_options]
 norecursedirs = ["bin", "lib", "include", "build"]
 
-[tool.ruff]
-target-version = "py38"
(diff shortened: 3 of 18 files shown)


woodruffw:
This repo is already formatted with `black`, which is declared as a dev-dep and is run via `make format`. I don't mind changing that to `ruff` instead, but I would strongly prefer that the switch is done by changing the dev dep + Makefile rather than adding a CI-only dependency that I'm almost certainly going to forget about 🙂

(Same goes for the spellchecker: no objection to having one, but please make it a dev dep.)

.github/workflows/tests.yml
```diff
@@ -9,6 +9,13 @@ on:
   workflow_dispatch:
 
 jobs:
+  codespell_and_ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: pip install --user codespell[toml] ruff
+      - run: codespell
+      - run: ruff check --output-format=github --target-version=py38 .
```
cclauss, line 18: Let's remove the automation and I will put that into a separate pull request.
```suggestion
```

Makefile
```diff
@@ -13,7 +13,9 @@ bootstrap: $(VENV)/bin/pip
 	$(VENV)/bin/pip install -e .[dev]
 
 format:
-	$(VENV)/bin/black .
+	$(VENV)/bin/codespell
+	$(VENV)/bin/ruff check
```
woodruffw, line 17: I believe this needs to be `ruff check --fix` to perform changes, right?

pyproject.toml
```diff
@@ -77,3 +78,6 @@ ignore_missing_imports = true
 
 [tool.pytest.ini_options]
 norecursedirs = ["bin", "lib", "include", "build"]
+
+[tool.ruff]
+target-version = "py38"
```
woodruffw, line 83: Our `requires-python` is `>= 3.7`, so this may cause `ruff` to enable fixes/changes that don't work with Python 3.7.

I believe you can drop this config section entirely and `ruff` *should* respect the top-level `requires-python` instead.

(Long term we'll want to remove 3.7 support, but we aim to retain it so long as the current `pip` version supports it.)

```suggestion
```

woodruffw:
Thanks @cclauss, two nitpicks but this LGTM overall :slightly_smiling_face:

woodruffw:
macOS with Python 3.7 is failing for unrelated reasons: https://github.com/actions/setup-python/issues/856

Makefile
cclauss: Some folks want ruff to autofix but others do not often because they are used to flake8 which never fixes.

cclauss, line 17:
```suggestion
	$(VENV)/bin/ruff check --fix
```

.github/workflows/tests.yml
```diff
@@ -18,6 +18,9 @@ jobs:
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
         os: ["macos-latest", "windows-latest", "ubuntu-latest"]
+        exclude:
+          - python-version: "3.7"
+            os: "macos-latest"
```
woodruffw, line 23: Let's just leave this out for now -- I'd rather merge this with a broken CI (or wait a little longer for GHA to fix it) than have a matrix exclude rule that I'll likely forget about :slightly_smiling_face:

cclauss, line 23:
```suggestion

```

woodruffw (approved):

woodruffw:
Thanks @cclauss! Please hold off on the CI changes for a bit -- I think what we'll want there is to have a `make lint` or similar that the CI then runs (using `setup-python`'s ability to cache deps).

cclauss:
I would highly recommend that you step up to `pre-commit` instead of `make`.

woodruffw:
I understand that people like `pre-commit`, but I'm not a huge fan of git hooks-based workflows: IME people frequently fail to configure them, and they require yet another piece of global tooling (`pip[x] install pre-commit`) that can break between Python versions.

That isn't to say that they're bad or that Makefiles are good (they aren't!), but I find them simpler to reason about. So I'm inclined to keep this repo in a "minimal" state with just a Makefile until Python's developer tooling circularities get resolved.

PR #335, opened 2024-05-08
Attempt to overwrite without race condition

### LINKED ISSUE #332: Race condition in FileCache that can result in empty body 200 response
I don't have a test case or suggested fix (yet), but documenting that this exists:

1. Poetry (unlike pip) uses cachecontrol's `FileCache` with no modifications.
2. If there is already a cached response, and two different processes (or threads too, probably) revalidate at the same time getting a 304, the first one goes through [`_cache_set`](https://github.com/psf/cachecontrol/blob/0e091c6524fae758b6c422a9e1aa16f99483e65c/cachecontrol/controller.py#L496-L497) and the second one can [find no body](https://github.com/psf/cachecontrol/blob/0e091c6524fae758b6c422a9e1aa16f99483e65c/cachecontrol/controller.py#L470-L474) (returning a response with empty body because `_secure_open_write` [deletes before creating](https://github.com/psf/cachecontrol/blob/0e091c6524fae758b6c422a9e1aa16f99483e65c/cachecontrol/caches/file_cache.py#L39-L52)).

Potential ideas:
* Check [for None body](https://github.com/psf/cachecontrol/blob/0e091c6524fae758b6c422a9e1aa16f99483e65c/cachecontrol/controller.py#L152-L155), and try harder (maybe `.get(locked=True)`.  Pro: simple, targeted change that only affects the rare case, Con: api change, deletion can still race
* Protect get (and delete) with the same lock that set uses, every time.  Pro: simple, Con: performance, race can still return mismatched body (similar to #324 with `SeparateBodyFileCache`)
* Stop using bespoke `mkstemp` that has a race, use `os.replace`.  Pro: less code, file isn't missing during the set, Con: commit history doesn't make it obvious why this exists, seems defensive against problems in world-writable cache dirs or maybe a windows issue?

thatch:
As explained in #332, there was previously a small window of time where the file is deleted before its new contents get written.  Because reads don't happen with the lock held, this resulted in an empty-body cache hit when it shouldn't.

Fixes #332

### DIFF
--- a/cachecontrol/caches/file_cache.py
+++ b/cachecontrol/caches/file_cache.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import hashlib
 import os
+import tempfile
 from textwrap import dedent
 from typing import IO, TYPE_CHECKING, Union
 from pathlib import Path
@@ -16,47 +17,6 @@ if TYPE_CHECKING:
     from datetime import datetime
 
     from filelock import BaseFileLock
-
-
-def _secure_open_write(filename: str, fmode: int) -> IO[bytes]:
-    # We only want to write to this file, so open it in write only mode
-    flags = os.O_WRONLY
-
-    # os.O_CREAT | os.O_EXCL will fail if the file already exists, so we only
-    #  will open *new* files.
-    # We specify this because we want to ensure that the mode we pass is the
-    # mode of the file.
-    flags |= os.O_CREAT | os.O_EXCL
-
-    # Do not follow symlinks to prevent someone from making a symlink that
-    # we follow and insecurely open a cache file.
-    if hasattr(os, "O_NOFOLLOW"):
-        flags |= os.O_NOFOLLOW
-
-    # On Windows we'll mark this file as binary
-    if hasattr(os, "O_BINARY"):
-        flags |= os.O_BINARY
-
-    # Before we open our file, we want to delete any existing file that is
-    # there
-    try:
-        os.remove(filename)
-    except OSError:
-        # The file must not exist already, so we can just skip ahead to opening
-        pass
-
-    # Open our file, the use of os.O_CREAT | os.O_EXCL will ensure that if a
-    # race condition happens between the os.remove and this line, that an
-    # error will be raised. Because we utilize a lockfile this should only
-    # happen if someone is attempting to attack us.
-    fd = os.open(filename, flags, fmode)
-    try:
-        return os.fdopen(fd, "wb")
-
-    except:
-        # An error occurred wrapping our FD in a file object
-        os.close(fd)
-        raise
 
 
 class _FileCacheMixin:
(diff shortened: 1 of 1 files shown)


woodruffw:
Thanks @thatch! I'll try and give this a review in the coming days.

thatch:
Any luck with testing?

woodruffw:
> Any luck with testing?

Sorry for the delay -- I was out for a bit, and I'm catching up now. I'll do some testing tonight.

woodruffw (approved):
I did some local testing (on Linux) and didn't hit any issues with this. 

I also went through the old `_secure_open_write` flags and confirmed that we shouldn't be missing anything important here -- the old flags should be irrelevant for our `mkstemp` use (since it defaults to binary mode and `O_NOFOLLOW` shouldn't be relevant for fresh temporary files). 

As such, this LGTM! I'd like @frostming to also do a comb over it though 🙂 

(I didn't test this locally on Windows, but my understanding is that `os.replace` is equally atomic there.)

thatch:
Ping @frostming

woodruffw:
@thatch I'll give @frostming another week or so to chime in, otherwise I'll merge and release this next week.

cachecontrol/caches/file_cache.py
```diff
@@ -122,15 +82,16 @@ def _write(self, path: str, data: bytes) -> None:
...
+        dirname = os.path.dirname(path)
+        os.makedirs(dirname, self.dirmode, exist_ok=True)
 
         with self.lock_class(path + ".lock"):
             # Write our actual file
-            with _secure_open_write(path, self.filemode) as fh:
-                fh.write(data)
+            (fd, name) = tempfile.mkstemp(dir=dirname)
+            os.write(fd, data)
+            os.close(fd)
```
graingert, line 92:
```suggestion
            (fd, name) = tempfile.mkstemp(dir=dirname)
            try:
                os.write(fd, data)
            finally:
                os.close(fd)
```

cachecontrol/caches/file_cache.py
```diff
@@ -122,15 +82,18 @@ def _write(self, path: str, data: bytes) -> None:
...
+        dirname = os.path.dirname(path)
+        os.makedirs(dirname, self.dirmode, exist_ok=True)
 
         with self.lock_class(path + ".lock"):
             # Write our actual file
-            with _secure_open_write(path, self.filemode) as fh:
-                fh.write(data)
+            (fd, name) = tempfile.mkstemp(dir=dirname)
+            try:
+                os.write(fd, data)
```
graingert, line 92: I don't think this is guaranteed to write all the data: https://docs.python.org/3/library/os.html#os.write

graingert (changes requested):

cachecontrol/caches/file_cache.py
woodruffw: Yep, I believe that's right. This is also true of the old code, since `fileobj.write(...)` can also perform a partial write.

I'll need to re-check, but I think the cases in which a partial write can occur here are pretty marginal when `fd` is a regular file: the main one that I remember is when the entire write would cause `ENOSPC`.

woodruffw (approved):
Looks good to me -- I think `_write` could potentially bail out harder if the underlying write fails rather than complete the `rename`, but that's worth thinking about separately.

PR #341, opened 2024-11-09
enable filterwarnings=['error'] and strict-makers, strict-config

graingert:


cachecontrol/filewrapper.py
```diff
@@ -117,3 +117,9 @@ def _safe_read(self, amt: int) -> bytes:
             self._close()
 
         return data
+
+    def close(self):
```
woodruffw, line 121: Not sure I follow this: there's a `_close` method too, and it seems to be disjoint in implementation from this one. Does this have an effect on the rest of the changes in this PR?

graingert: FileWrapper is assigned as a `HTTPResponse._fp` which if it as a close method gets called by HTTPResponse.close` which if the request is streaming needs to be called by using the response as a context manager

graingert: The windows issues are new and exciting, I'll have to investigate them later

graingert: The `_close` is disjoint because it's called when the fp is exhausted and the buf has finished being written to, CallbackFileWrapper.close is called when the consumer of the response failed reading the body in some way, if the response was streaming

graingert: > Does this have an effect on the rest of the changes in this PR?

It's needed otherwise you get a ResourceWarning that's propagated as an unraisable exception

graingert, line 121:
```suggestion
    def close(self) -> None:
```

tests/test_regressions.py
```diff
@@ -23,11 +23,12 @@ def test_file_cache_recognizes_consumed_file_handle(self, url):
...
     s = CallbackFileWrapper(None, None)
-    # normal behavior:
-    with pytest.raises(AttributeError):
-        s.x
-
-    # this previously had caused an infinite recursion
-    vars(s).clear()  # gc does this.
-    with pytest.raises(AttributeError):
-        s.x
+    with s._CallbackFileWrapper__buf:
```
graingert, line 26: it seems NamedTemporaryFile only issues a ResourceWarning on windows https://github.com/python/cpython/issues/126639

cachecontrol/filewrapper.py
graingert: I resolved the windows issues - turns out there's a bug in cpython! I think I've found a cpython bug every time I've added filterwarnings=["error"] to a project!

graingert:
@woodruffw can you take a look at this please?

woodruffw:
Yep, I'll do another review pass tonight.

PR #344, opened 2024-12-13
fix: response leak when using caching adapter and streaming

Tasssadar:
Currently, this library breaks `requests` behavior, where the finalizer will close response if user forges to call close. Use WeakRefs to prevent this.

### DIFF
--- a/cachecontrol/adapter.py
+++ b/cachecontrol/adapter.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import functools
 import types
+import weakref
 import zlib
 from typing import TYPE_CHECKING, Any, Collection, Mapping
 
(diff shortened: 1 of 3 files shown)


cachecontrol/adapter.py
```diff
@@ -128,19 +129,25 @@ def build_response(  # type: ignore[override]
...
+                        self = weak_self()
+                        if self is None:
+                            return
+
+                        super_update_chunk_length(self)
                         if self.chunk_left == 0:
                             self._fp._close()  # type: ignore[union-attr]
 
-                    response._update_chunk_length = types.MethodType(  # type: ignore[method-assign]
-                        _update_chunk_length, response
```
Tasssadar, line 143: When you assign a method like this, you're creating a circular reference to `self`, because when you normally call `response._update_chunk_length`, it actually creates the bound method on-the-fly, it's not set there all the time.

This SO response helped me understand it: https://stackoverflow.com/a/26158130

cachecontrol/controller.py
```diff
@@ -323,7 +324,7 @@ def _cache_set(
     def cache_response(
         self,
         request: PreparedRequest,
-        response: HTTPResponse,
+        response_or_ref: HTTPResponse | weakref.ReferenceType[HTTPResponse],
```
frostming, line 327: Is there any chance it would receive a weakref?

Tasssadar: Yes, the code I changed in adapter.py on line 132 puts it there. This function looks like it is part of public API, so I didn't change it to only accept weakref, only extended it.

I just added a comment below to make clearer why it can be None, 
```
# The weakref can be None only in case the user used streamed request
# and did not consume or close it, and holds no reference to requests.Response.
# In such case, we don't want to cache the response.```

frostming (approved):
Good catch and fix. I would like to accept this. And I will wait for the inputs from @woodruffw

woodruffw (approved):
Thanks, makes sense to me as well. I appreciate the additional unit test!

racinmat:
Hi, when can we expect this being released to pypi?

woodruffw:
> Hi, when can we expect this being released to pypi?

Assuming nothing else comes up, I'll cut a release in the next day or so.

PR #345, opened 2024-12-16
ci: harden workflows

woodruffw:
This hash-pins all of our workflows and fixes some (low-impact) zizmor findings.

It also adds a `zizmor` workflow that'll run on pushes and PRs to prevent issues from sneaking in.


woodruffw:
CC @frostming for approval (since this is my tool + we should really have two-person signoffs here, but I don't have this ability to enable them 😅)

PR #405, opened 2026-01-06
cache temporary redirects with explicit caching directives

### LINKED ISSUE #384: HTTP 302 Found responses with Cache Headers not being cached

cburroughs:
RFC 9111 §3 - Storing Responses in Caches, describes the conditions under which a cache may store a response.  Notably all final (aka not 1xx) status codes are *potentially* cacheable, subject to the other listed conditions.

In brief a response must have either an explicit header (such as a `max-age` directive) OR have a heuristically cacheable status code. Which status codes are heuristically cacheable is helpfully summarized in RFC 9110 §15.1 - Overview of Status Codes: 200, 203, 204, 206, 300, 301, 308, 404, 405, 410, 414, and 501)

CacheController has historically maintained a more conservative list of cacheable_status_codes that are a subset of the heuristically cacheable ones.

This diff makes related changes:
 * Temporary redirects are now cached by default.  Besides being correct in the general case, temporary redirects are used by several Python Registries as discussed in #384.
 * Because temporary redirects are not heuristically cacheable, add a check to only cache them when explicit headers are present.  This means that regardless of which status_codes are passed to CacheController, the HTTP specification is still followed.
 * As defensive future proofing, make sure the response is final before much else is done.

fixes #384

### DIFF
--- a/cachecontrol/controller.py
+++ b/cachecontrol/controller.py
@@ -18,6 +18,7 @@ from typing import TYPE_CHECKING, Collection, Mapping
 
 from requests.structures import CaseInsensitiveDict
 
+from cachecontrol.heuristics import HEURISTICALLY_CACHEABLE_STATUSES
 from cachecontrol.cache import DictCache, SeparateBodyBaseCache
 from cachecontrol.serialize import Serializer
 
@@ -32,6 +33,17 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 URI = re.compile(r"^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?")
+
+NEVER_CACHE_STATUSES = (
+    # Per https://www.rfc-editor.org/rfc/rfc9111.html#section-3 the status
+    # code must be final
+    100,
+    101,
+    # From httplib2: Don't cache 206's since we aren't going to
+    #                handle byte range requests
+    206,
+)
+
 
 PERMANENT_REDIRECT_STATUSES = (301, 308)
 
@@ -60,7 +72,20 @@ class CacheController:
         self.cache = DictCache() if cache is None else cache
         self.cache_etags = cache_etags
         self.serializer = serializer or Serializer()
-        self.cacheable_status_codes = status_codes or (200, 203, 300, 301, 308)
+        # Per https://www.rfc-editor.org/rfc/rfc9111.html#section-3-2.7.1 all
+        # all final response codes are potentially cacheable, subject to the
+        # other conditions.  CacheController conservatively only caches common
+        # ones codes.  For example, even with a max-age set, a 500 Internal
+        # Server Error will not be cached
+        self.cacheable_status_codes = status_codes or (
+            200,
+            203,
+            300,
+            301,
+            302,
+            307,
+            308,
+        )
 
     @classmethod
     def _urlnorm(cls, uri: str) -> str:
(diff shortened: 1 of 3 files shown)


woodruffw:
Sorry for the delay @cburroughs, I'll try and set aside some time to review this in the coming days.

cburroughs:
I'll try the opposite argument this time and say I'd love to get this landed before the upcoming Pip 26.2 <https://github.com/pypa/pip/issues/14067> ;-)

  (Although I know that with Pip's vendor-syncing cadence that might mean Pip doesn't pick this up until 26.3)

cachecontrol/controller.py
```diff
@@ -60,7 +72,20 @@ def __init__(
         self.cache = DictCache() if cache is None else cache
         self.cache_etags = cache_etags
         self.serializer = serializer or Serializer()
-        self.cacheable_status_codes = status_codes or (200, 203, 300, 301, 308)
+        # Per https://www.rfc-editor.org/rfc/rfc9111.html#section-3-2.7.1 all
+        # all final response codes are potentially cacheable, subject to the
+        # other conditions.  CacheController conservatively only caches common
+        # ones codes.  For example, even with a max-age set, a 500 Internal
```
woodruffw: Q: what does "common ones" mean here?

cachecontrol/controller.py
```diff
@@ -33,6 +34,17 @@
 
 URI = re.compile(r"^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?")
 
+NEVER_CACHE_STATUSES = (
```
woodruffw: Nit: let's make this and the other constants introduced by this PR private, since we're not committing to them as new public APIs 🙂 


```suggestion
_NEVER_CACHE_STATUSES = (
```

woodruffw: (We do already have some public constants, which is unfortunate. So I'm not 100% opposed to making these public if you have an argument for it, but by default I'd prefer not to make the public API surface any bigger.)

woodruffw (commented):
Thanks @cburroughs, sorry for my tardiness here. This looks great to me overall, just one question and one nitpick 🙂

cachecontrol/controller.py
cburroughs: * I `_` prefixed `NEVER_CACHE_STATUSES`
 * For `HEURISTICALLY_CACHEABLE_STATUSES` this is replacing `LastModified.cacheable_by_default_statuses` which didn't have a `_` so I think that very slight leads towards leaving `HEURISTICALLY_CACHEABLE_STATUSES` without one.  Happy to prefix that one as well if you prefer.

cachecontrol/controller.py
cburroughs: I've rephrased the comment to -- I hope! -- be clearer.  The existing name `cacheable_status_codes` is somewhat ambiguous.   Is is "the status codes that the `CacheController` class will consider cacheing", *not* "which status codes are cacheable per the specification".  The original intent as I understand it was to be conservative and not get bogged down in edge cases and quirky implimentations.  https://mnot.net/blog/2017/browser-caching is somewhat old at this point but is one example of Firefox/Safari/Chrome all taking different stances on questions like "cacheing a 500 is allowed, but do we think that is really what somewhat meant to do or are they shooting themselves in the foot?"

My motivation for this change was that I'm working with a package index built around redirecting from a web a web server to cloud provider blob storage (ex: S3).  I think this a reasonable and common design and I'm trying to make the smallest relaxation to the existing conservative rules that supports that.  I think it's less likely that anyone has designed a system around the cacheable of something like  "414 URI Too Long".  (But if they did this would be the place in the code to look.)

PR #425, opened 2026-05-19
Avoid a race-condition on ETag validation.

### LINKED ISSUE #424: Race condition in send()
I spotted this while using `tox -vvv -p all`, so via `pip`, since `pip==26.1`.

The issue with `tox -p all` is that it can run multiple versions of Python (in my config from 3.10 to 3.14), in parallel, I bet some with zstd some without.

TL;DR:

- pip wants to query /simple/setuptools/
- Cache is checked, an etag is found in the cache.
- The HTTP request with an If-None-Match is sent.
- In the meantime the cache gets modified, from `Accept-Encoding: gzip, deflate` to `Accept-Encoding: gzip, deflate, zstd` or the other way around, by another process.
- The HTTP response arrives, it's a `304 Not Modified`, the cache has to be updated, but this time it does no longer loads, and the 304 itself is returned.

In details:

It all starts in those two lines of `adapter.send()`:

            request.headers.update(self.controller.conditional_headers(request))
        resp = super().send(request, stream, timeout, verify, cert, proxies)

Here `conditional_headers` does a `_load_from_cache` which can succeed, let say it succeed.

Then `super().send()` will eventually call `adapter.build_response()` with a `304 Not Modified` so `build_response` calls `self.controller.update_cached_response` which runs another `_load_from_cache`, which can fail if the cache was modified by another process. Let's say it fails it runs this code:

        if not cached_response:
            # we didn't have a cached response
            return response

So the 304 response is returned, up to the user. In case of `pip` an exception is raised because the answer don't have a `Content-Type`. But `pip` or not, the user just did a `get` with no `If-None-Match` so he does not expect a `304`, either a cache hit (hopefully 200 OK) or a cache miss (hopefully 200 OK) but never a `304`.

# The chronology with two processes

It think it looks like this, `[1]` and `[2]` denoting two distinct processes:

- [1] runs a `requests.get("https://pypi.org/simple/setuptools/")`
- [2] runs a `requests.get("https://pypi.org/simple/setuptools/")`
- [1] Read the cache, extract an ETag from the cache
- [2] Read the cache, extract an ETag from the cache
- [1] Sends the HTTP request with If-None-Match
- [2] Sends the HTTP request with If-None-Match
- [1] Gets a response, `adapter.build_response` calls `self.controller.update_cached_response()`, changing the headers on disk. The `Accept-Encoding` in my case.
- [1] Returns the cached response.
- [2] Gets a response, `adapter.build_response` calls `self.controller.update_cached_response()` which calls `_load_from_cache` which refuses to load because the headers has changed (`Vary` tells the answer varies on `Accept-Encoding` and the `Accept-Encoding` changed), `update_cached_response` returns the 304 response. `build_response` returns the 304 response. The user gets a `304`, which is unexpected.

JulienPalard:
For example, in case multiple Python versions are using the same cache and some are <3.14 and other >=3.14, the Accept-Encoding header will vary between `gzip, deflate` and `gzip, deflate, zstd`.

If two processes, say p1 (with zstd support) and p2 (without zstd support) do two requests to the same URL, with an already cached result (for the without zstd support request) it could lead to the following race condition:

- P1 reads the cache, fails because Accept-Encoding does not match
- P1 sends a normal request
- P2 reads the cache, succeeds, but needs validation
- P2 sends a validation request (with If-None-Match)
- P1 receives a response for its Accept-Encoding: gzip, deflate, zstd
- P1 updates the cache with it
- P1 returns the 200 OK to the user
- P2 receives a 304 Not Modified response
- P2 wants to return from the cache, but fails to read it (wrong Accept-Encoding!)
- P2 fallbacks to returning a 304 Not Modified response  to the user :(

Closes #424

### DIFF
--- a/cachecontrol/controller.py
+++ b/cachecontrol/controller.py
@@ -61,6 +61,7 @@ class CacheController:
         self.cache_etags = cache_etags
         self.serializer = serializer or Serializer()
         self.cacheable_status_codes = status_codes or (200, 203, 300, 301, 308)
+        self._request_to_cache_response = weakref.WeakKeyDictionary()
 
     @classmethod
     def _urlnorm(cls, uri: str) -> str:
@@ -281,6 +282,10 @@ class CacheController:
         new_headers = {}
 
         if resp:
+            # Save the cached response associated with the request so
+            # if the server returns a 304 we can use exactly this response.
+            # (usefull in case the cache gets altered in the meantime)
+            self._request_to_cache_response[request] = resp
             headers: CaseInsensitiveDict[str] = CaseInsensitiveDict(resp.headers)
 
             if "etag" in headers:
@@ -479,7 +484,7 @@ class CacheController:
         """
         assert request.url is not None
         cache_url = self.cache_url(request.url)
-        cached_response = self._load_from_cache(request)
+        cached_response = self._request_to_cache_response.get(request) or self._load_from_cache(request)
 
         if not cached_response:
             # we didn't have a cached response


JulienPalard:
As a nice side effect this save one disk read in case the server answers a 304 to a validation request.

But I'm not proud of my edit of `conditional_headers`, it previously was kind of pure, it now alters a "global state". It would have been better to have this to the adapter. But having the state modification to the adapter is worse: it can't cleanly call `self.controller._load_from_cache` (private Controller method), and it alters the prototype of `conditional_headers` (passing it the cached response).

So if someone with a better [theory of the project](https://gwern.net/doc/cs/algorithm/1985-naur.pdf) than me has a better idea, don't hesitate. After all that's just the first time I read `cachecontrol` code.

PR #449, opened 2026-07-20
fix(controller): Recover from invalid dates

DarkaMaul:
Tiny hardening change that prevents a malformed Date header from crashing the controller.


woodruffw (approved):
Seems fine to me, thanks @DarkaMaul!

PR #465, opened 2026-08-28
lock once for metadata and body in separate body cache

ogalland-tob:
`SeparateBodyFileCache` replaces an entry's metadata and body files separately. A concurrent read can combine metadata from one update with the body from another and return the mixed-generation entry as a valid cached response.

This PR introduces atomic `get/set_with_body` operations to prevent the race condition, as well as relevant unit tests.

### DIFF
--- a/cachecontrol/cache.py
+++ b/cachecontrol/cache.py
@@ -57,13 +57,36 @@ class SeparateBodyBaseCache(BaseCache):
     In this variant, the body is not stored mixed in with the metadata, but is
     passed in (as a bytes-like object) in a separate call to ``set_body()``.
 
-    That is, the expected interaction pattern is::
+    Cache implementations can override ``set_with_body()`` and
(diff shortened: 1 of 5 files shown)


igormishsky:
We’re hitting this issue in CI with Poetry 2.4.1: parallel installs intermittently fail to resolve setuptools.
Our [standalone reproducer](https://github.com/igormishsky/cachecontrol/blob/3af1c3c1e03425793d67a7d0a4799b7f4ed01607/scripts/reproduce_poetry_cache_race.py) failed 20/20 times with CacheControl 0.14.4 and passed 20/20 with this PR’s commit 7df3205, using unmodified Poetry with caching enabled.
Could a maintainer approve CI and review this fix?

I also opened [Poetry issue #11048](https://github.com/python-poetry/poetry/issues/11048) to track the impact on Poetry, including reproduction steps and test results.

PR #466, opened 2026-08-28
don't lower case-sensitive user info embedded in URLs

ogalland-tob:
CacheControl lowercases URL user information when constructing cache keys. Requests preserves its case and generates different Basic Authorization headers, allowing a response for one credential spelling to satisfy another.

This PR prevents the lowering of userinfo in URL and introduces a unit test for this case.

### DIFF
--- a/cachecontrol/controller.py
+++ b/cachecontrol/controller.py
@@ -70,7 +70,8 @@ class CacheController:
             raise Exception("Only absolute URIs are allowed. uri = %s" % uri)
 
         scheme = scheme.lower()
-        authority = authority.lower()
+        userinfo, separator, host = authority.rpartition("@")
+        authority = f"{userinfo}{separator}{host.lower()}"
 
         if not path:
             path = "/"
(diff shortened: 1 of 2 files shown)


cachecontrol/controller.py
```diff
@@ -70,7 +70,8 @@ def _urlnorm(cls, uri: str) -> str:
             raise Exception("Only absolute URIs are allowed. uri = %s" % uri)
 
         scheme = scheme.lower()
-        authority = authority.lower()
+        userinfo, separator, host = authority.rpartition("@")
```
woodruffw, line 73: I'm a little worried about this being conformant -- does RFC 3986 or similar guarantee that partitioning on the last `@` is correct?

ogalland-tob: Good point, I was assuming there could only be one `@` in the authority, but I haven't checked against the RFC. Will take a look now.

ogalland-tob: As per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), `authority   = [ userinfo "@" ] host [ ":" port ]`. afaiu `@` is excluded from all of these symbols. The `@` character is only included in `gen-delims`, which does not show up in any of `userinfo`, `host`, `port`.

woodruffw (approved):

PR #467, opened 2026-08-28
parse cache-control directive case-insensitively

ogalland-tob:
CacheControl matches Cache-Control directive names case-sensitively, although RFC 9111 requires case-insensitive matching. Valid directives such as No-Store are therefore ignored.

This PR normalizes directive name cases and introduces a test case.


woodruffw (approved):
Thanks, good catch!

PR #468, opened 2026-08-28
apply freshness checks to redirects with explicit expiration

ogalland-tob:
CacheControl treats stored 301 and 308 responses as permanently reusable. It returns them before evaluating response freshness or cache-control directives.

This is a bit tricky, but RFC 9111 section 4.2.2 states:

> A cache MUST NOT use heuristics to determine freshness when an explicit expiration time is present in the stored response. Because of the requirements in [Section 3](https://www.rfc-editor.org/info/rfc9111/#response.cacheability), heuristics can only be used on responses without explicit freshness whose status codes are defined as "heuristically cacheable" (e.g., see [Section 15.1](https://www.rfc-editor.org/info/rfc9110/#section-15.1) of [[HTTP](https://www.rfc-editor.org/info/rfc9111/#HTTP)]) and on responses without explicit freshness that have been marked as explicitly cacheable (e.g., with a public response directive).

If we take cachecontrol's heuristic to be "the response is permanently fresh" then (afaiu) this would mean that redirect responses with explicit expiration dates _can_ expire and thus should be checked for freshness (i.e., the explicit expiration takes precedence over the heuristic).

This PR applies the usual freshness checks to redirects, and introduces a fallback to still treat them as permanently fresh if the response is not expired. It also introduces relevant unit tests.

### DIFF
--- a/cachecontrol/controller.py
+++ b/cachecontrol/controller.py
@@ -190,23 +190,11 @@ class CacheController:
         if not resp:
             return False
 
-        # If we have a cached permanent redirect, return it immediately. We
-        # don't need to test our response for other headers b/c it is
-        # intrinsically "cacheable" as it is Permanent.
-        #
-        # See:
-        #   https://tools.ietf.org/html/rfc7231#section-6.4.2
-        #
-        # Client can try to refresh the value by repeating the request
-        # with cache busting headers as usual (ie no-cache).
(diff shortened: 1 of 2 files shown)


woodruffw (approved):

woodruffw:
Thank you @ogalland-tob! Seems right to me.

PR #469, opened 2026-08-28
don't omit CRLF when those are the only bytes the in the chunk

ogalland-tob:
`CallbackFileWrapper._safe_read()` omits a legitimate two-byte CRLF chunk from the cached response body. If a response contains a chunk consisting exclusively of CRLF, it is omitted from the cached response.

This PR fixes the issue by keeping track of remaining bytes in `CallbackFileWrapper` and only omitting the trailing CRLF. It also introduces a relevant unit test.

### DIFF
--- a/cachecontrol/adapter.py
+++ b/cachecontrol/adapter.py
@@ -144,6 +144,10 @@ class CacheControlAdapter(HTTPAdapter):
                         super_update_chunk_length(self)
                         if self.chunk_left == 0:
                             self._fp._close()  # type: ignore[union-attr]
+                        elif self.chunk_left is not None:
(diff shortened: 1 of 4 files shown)


woodruffw (approved):
Thanks @ogalland-tob!

Making sure I understand: this happens when we're reading chunks of an HTTP response, and previously we'd accidentally discard a CRLF if it happened to be the only contents of a chunk?

ogalland-tob:
Yes, this would cause chunks made up entirely of CRLF to be dropped.

woodruffw:
Thanks @ogalland-tob!

